PR #3, opened 2014-12-26
Use @staticmethod where appropriate

### LINKED ISSUE #2: [Refactor] Static methods in DBO
In the Mongo DBO (`ivre/ivre/db/mongo.py`), some methods require a `MongoDB` instance, but they do not use it.

For instance:
* `getid`
* `serialize`
* `distinct`
* `str2id`
* `_flt_and`
* `flt_or`
* ...

Since these methods are related to MongoDB specificities but independent of the current instance, they could be implemented as static methods.

Does it appear legitimate ?

p-l-:
This fixes #2. Thanks to Camille 'Kamoul0x' Mougey for reporting this.

### DIFF
--- a/ivre/db/__init__.py
+++ b/ivre/db/__init__.py
@@ -81,7 +81,8 @@ class DB(object):
         """
         return reduce(self._flt_and, args)
 
-    def _flt_and(self, cond1, cond2):
+    @staticmethod
+    def _flt_and(cond1, cond2):
         """Returns a condition that is true iff both `cond1` and
         `cond2` are true.
 
@@ -98,7 +99,8 @@ class DB(object):
         """
         return reduce(self._flt_or, args)
 
-    def _flt_or(self, cond1, cond2):
+    @staticmethod
+    def _flt_or(cond1, cond2):
         """Returns a condition that is true iff either `cond1` or
         `cond2` is true.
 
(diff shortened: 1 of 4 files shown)


p-l-:
I'm interested in cleaning up these functions too.

Suppose we make `searchjavaua` a `@staticmethod`. How can we write the `self.searchua()` call?

If we decide to make it a `@classmethod` instead, because we can write `cls.searchua()`, what is the advantage over letting a regular method and using `self`?

commial:
To my mind, there are three distinct methods' types in Python. That is to say, `staticmethod`, `classmethod` and standard methods.

The method type is part of the method signature. So, imposing a method type is a way to give some hints about the function behavior.

For instance:
* A standard method, which take `self` as input, may depend on the current instance;
* A `@classmethod` has to depend only on the current class;
* A `@staticmethod` will never change. Even if you inherits and override a static method, the call will remain on the first method.

That way, when one defines a `@classmethod`, he indicates that this method and its dependencies may be override. That is not the case when defining a `@staticmethod`.

To sum up, in our example, one will write `searchjavaua` as:
* a `@staticmethod`, which will always call `DB.searchua()` and has to be re-implemented by children if the behavior has to change;
* a `@classmethod`, which will depends on the implementation of `searchua` in the current class;
* a standard method, which will depends on the current instance (for example, it depends on the last two queries).

I'm not sure to be very clear. I think that, in our case, the signature you expect for `searchjavaua` is a `@classmethod`.

p-l-:
OK so that's a distinct work to do that can be done separately. I'm merging this one for now, and will do the rest later.

PR #10, opened 2015-01-13
Python 3 compatibility

p-l-:
**Do not merge for now / this is open for comments only**

This PR supports Python 3 compatibility.

This code **must** be compatible with Python versions from 2.6 to 3.4 (or at least 2.6, 2.7 and 3.4).


p-l-:
After some work on Python 2 & 3 support, I do not think the effort is worth it.

PR #13, opened 2015-02-10
Support for screenshots

p-l-:
This adds support for screenshots (MongoDB, Web UI) taken with external tools.

### DIFF
--- a/ivre/db/mongo.py
+++ b/ivre/db/mongo.py
@@ -408,10 +408,14 @@ have no effect if it is not expected)."""
         if url == "field":
             return port.get('screendata')
 
-    def setscreenshot(self, host, port, data, archives=False,
-                      overwrite=False):
+    def setscreenshot(self, host, port, data, protocol='tcp',
+                      archives=False, overwrite=False):
         """Sets the content of a port's screenshot."""
-        port = [p for p in host['ports'] if p['port'] == port][0]
+        try:
+            port = [p for p in host.get('ports', [])
+                    if p['port'] == port and p['protocol'] == protocol][0]
+        except IndexError:
+            raise KeyError("Port %s/%d does not exist" % (protocol, port))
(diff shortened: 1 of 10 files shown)


ivre/db/mongo.py
```diff
@@ -390,6 +400,43 @@ def getscan(self, scanid, archive=False, **kargs):
...
 
+    def getscreenshot(self, port):
+        """Returns the content of a port's screenshot."""
+        url = port.get('screenshot')
+        if url is None:
+            return None
+        if url == "field":
+            return port.get('screendata')
+
+    def setscreenshot(self, host, port, data, archives=False,
```
commial: Can you add a `protocol` parameter, maybe with a default value to `TCP`?

ivre/db/mongo.py
```diff
@@ -390,6 +400,43 @@ def getscan(self, scanid, archive=False, **kargs):
...
+                      overwrite=False):
+        """Sets the content of a port's screenshot."""
+        port = [p for p in host['ports'] if p['port'] == port][0]
```
commial: You should raise a proper exception if the port doesn't exist

ivre/db/mongo.py
```diff
@@ -390,6 +400,43 @@ def getscan(self, scanid, archive=False, **kargs):
...
+        port['screenshot'] = "field"
+        port['screendata'] = bson.Binary(data)
+        self.db[
+            self.colname_oldhosts if archives else self.colname_hosts
+        ].update({"_id": host['_id']}, {"$set": {'ports': host['ports']}})
+
+    def removescreenshot(self, host, port=None, archives=False):
+        """Removes screenshots"""
+        changed = False
+        ports = host.get('ports')
```
commial: You probably want to use `.get('ports', [])`?

PR #16, opened 2015-03-06
Add functions to look for open SMB shares

p-l-:
This code uses the structured data from `smb-enum-shares.nse` Nmap script.

This requires a recent version of Nmap **and either** to reload the scans from the Nmap XML files **or** to use a script to adapt the results from your DB, like:

```python
def update_smb_enum_shares(archive=False):
    cursor = ivre.db.db.nmap.get(
        ivre.db.db.nmap.searchscript(name='smb-enum-shares', host=True),
        fields=["_id", "scripts"],
        archive=archive,
    )
    collection = ivre.db.db.nmap.colname_oldhosts if archive else ivre.db.db.nmap.colname_hosts
    for host in cursor:
        changed = False
        for script in host['scripts']:
            if script['id'] == 'smb-enum-shares' and 'smb-enum-shares' in script:
                script['smb-enum-shares'] = ivre.xmlnmap.change_smb_enum_shares(script['smb-enum-shares'])
                changed = True
        if changed:
            ivre.db.db.nmap.db[collection].update(
                {'_id': host['_id']},
                {'$set': {
                    'scripts': host['scripts'],
                }}
            )

update_smb_enum_shares()
update_smb_enum_shares(archive=True)
```

### DIFF
--- a/ivre/db/mongo.py
+++ b/ivre/db/mongo.py
@@ -308,6 +308,7 @@ class MongoDBNmap(MongoDB, DBNmap):
 
     content_handler = xmlnmap.Nmap2Mongo
     needunwind = ["categories", "ports", "ports.scripts", "scripts",
+                  "scripts.smb-enum-shares.shares",
                   "extraports.filtered", "traces", "traces.hops",
                   "os.osmatch", "os.osclass", "hostnames",
                   "hostnames.domains"]
@@ -903,6 +904,35 @@ have no effect if it is not expected)."""
                           'http-vlcstreamer-ls', 'nfs-ls', 'smb-ls']},
             output=fname,
         )
+
+    def searchsmbshares(self, access='', hidden=None):
+        """Filter SMB shares with anonymous `access` (default: either
+        read or write, accepted values 'r', 'w', 'rw').
+
+        If `hidden` is set to `True`, look for hidden shares, for
+        non-hidden if set to `False` and for both if set to `None`
+        (this is the default).
+
+        """
+        access = {
+            '': re.compile('^(READ|WRITE)'),
+            'r': re.compile('^READ(/|$)'),
+            'w': re.compile('(^|/)WRITE$'),
+            'rw': 'READ/WRITE',
+            'wr': 'READ/WRITE',
+        }[access.lower()]
+        share_type = {
+            None: re.compile('^STYPE_DISKTREE(_HIDDEN)?'),
+            True: 'STYPE_DISKTREE_HIDDEN',
+            False: 'STYPE_DISKTREE',
+        }[hidden]
+        return self.searchscript(
+            name='smb-enum-shares',
+            values={'shares': {'$elemMatch': {
+                'Anonymous access': access,
+                'Type': share_type,
+            }}},
+            host=True)
 
     def searchhttptitle(self, title):
         return self.searchscript(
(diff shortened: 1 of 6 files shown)


ivre/db/mongo.py
```diff
@@ -904,6 +904,28 @@ def searchfile(self, fname):
             output=fname,
         )
 
+    def searchsmbshares(self, access=None):
+        """Filter SMB shares with anonymous `access` (default: either
+        read or write, accepted values 'r', 'w', 'rw')
+
+        """
+        if access is None:
```
commial: Why don't you set the default value of `access` to `''` ?

ivre/db/mongo.py
```diff
@@ -904,6 +904,28 @@ def searchfile(self, fname):
...
+            access = ''
+        access = access.lower()
+        if 'w' in access:
```
commial: This code snippet is quite hard to understand. You could set the default value of `access` to `re.compile('^(READ|WRITE)')` and then apply a simpler test.
Or use a dictionnary with `access` as a key.

ivre/xmlnmap.py
```diff
@@ -470,22 +491,27 @@ def endElement(self, name):
...
-                            if infos:
-                                self._curscript[infokey] = infos
-                elif hasattr(infos, "__call__"):
-                    infos = infos(self._curscript)
-                    if infos is not None:
-                        self._curscript[infokey] = infos
+            elif infokey != 'infos':
+                if infokey in ADD_TABLE_ELEMS:
+                    infos = ADD_TABLE_ELEMS[infokey]
+                    if type(infos) is utils.REGEXP_T:
```
commial: Generally, it is better to use `isinstance` instead of `type(..) is ..`.
(Point `Object type comparisons should always use isinstance() instead of comparing types directly. ` in [PEP08](https://www.python.org/dev/peps/pep-0008/))

ivre/xmlnmap.py
```diff
@@ -470,22 +491,27 @@ def endElement(self, name):
...
+                        infos = infos.search(self._curscript.get('output', ''))
+                        if infos is not None:
+                            infosdict = infos.groupdict()
+                            if infosdict:
+                                self._curscript[infokey] = infosdict
+                            else:
+                                infos = list(infos.groups())
+                                if infos:
```
commial: You use three time `infos` for different, imbricated, elements. It may be a good thing to use different variable names.

ivre/db/mongo.py
p-l-: Because I'm an idiot!

ivre/db/mongo.py
p-l-: Nice one! I was looking for a good way to implement that but couldn't find one. The dictionnary is an excellent idea!

ivre/xmlnmap.py
p-l-: Agreed, but that's code from previous version (it appears in the diff because the indentation has changed due to a condition factorisation w/ new code).

ivre/db/mongo.py
```diff
@@ -904,6 +905,31 @@ def searchfile(self, fname):
             output=fname,
         )
 
+    def searchsmbshares(self, access='', hidden=None):
+        """Filter SMB shares with anonymous `access` (default: either
```
commial: What's the purpose of the parameter `hidden` ? You docstring for `access` but not for it :)

p-l-: You're right.

PR #24, opened 2015-03-23
Add Travis CI framework

p-l-:
This also fixes the way tests are done.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,19 @@ install:
   - pip install .
   - pip install -r requirements.txt
 
-services: mongodb
+# We need MongoDB 2.6
+# https://github.com/travis-ci/travis-ci/issues/2246
+#services: mongodb
 
 before_script:
+  # install MongoDB and start it
+  - sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 7F0CEB10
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -2,6 +2,8 @@ This file is part of IVRE.
 
 Copyright 2011 - 2014 [Pierre LALET](mailto:pierre.lalet@cea.fr)
 
+[![Build Status](https://travis-ci.org/cea-sec/ivre.svg?branch=feature-travis-tests)](https://travis-ci.org/cea-sec/ivre)
```
commial: You may want to use `branch=master` ?

.travis.yml
```diff
@@ -0,0 +1,41 @@
+language: python
+
+python:
+  - 2.7
```
commial: Here, you can add `2.6` to ensure the backward compatibility

README.md
p-l-: Yep, fixed that (later)

.travis.yml
p-l-: Unfortunately, the tests do not work with 2.6 (because unittest lacks some features in 2.6 it has in 2.7). Nothing impossible to do, but that would require work. And I hope I'll be able to drop 2.6 support one day ;-)

PR #27, opened 2015-04-08
Docker: add an ivre/web image using Apache

p-l-:
Default image uses Nginx.


p-l-:
This PR also forces pymongo's version since the API from version 3.0 (which is installed by default now) is not compatible with 2.7.

PR #31, opened 2015-05-01
Parsing CPE (Common Platform Enumeration) strings

fmonjalet:
This PR adds a basic support of CPE (Common Platform Enumeration) returned by some nmap components (osmatch, scripts or port information).

- `xmlnmap.py` now parses CPE information: any cpe is added to to the element that found it, for example:
```
[...]
    ports: [
            {
                "service_method": "probed",
                "state_reason": "syn-ack",
                "state_reason_ttl": 119,
                "protocol": "tcp",
                "service_name": "msrpc",
                "state_state": "open",
                "cpes": [
                    "cpe:/o:microsoft:windows"
                ],
                "service_product": "Microsoft Windows RPC",
                "service_conf": 10,
                "service_ostype": "Windows",
                "port": 135
            },
[...]
            "osmatch": [
                {
                    "line": "59470",
                    "cpes": [
                        "cpe:/o:microsoft:windows_7::-",
                        "cpe:/o:microsoft:windows_7::sp1",
                        "cpe:/o:microsoft:windows_server_2008::sp1",
                        "cpe:/o:microsoft:windows_8"
                    ],
                    "name": "Microsoft Windows 7 SP0 - SP1, Windows Server 2008 SP1, or Windows 8",
                    "accuracy": "100"
                }
            ],
[...]
```

This information is also parsed and aggregated at the host level:
```
        "cpes": [
            {
                "product": "windows_server_2008",
                "vendor": "microsoft",
                "origins": [
                    "os.osmatch.line:59470"
                ],
                "value": "cpe:/o:microsoft:windows_server_2008::sp1",
                "components": ":sp1",
                "type": "o"
            },
            {
                "product": "windows",
                "vendor": "microsoft",
                "origins": [
                    "ports.port:135",
                    "ports.port:2103",
                    "ports.port:2105",
                    "ports.port:2107",
                    "ports.port:49152",
                    "ports.port:49153",
                    "ports.port:49154",
                    "ports.port:49155",
                    "ports.port:49156"
                ],
                "value": "cpe:/o:microsoft:windows",
                "components": "",
                "type": "o"
            },
[...]
        ]
```
This parsing is very simple, and do not cover the full CPE specification, only examples found in nmap xml scan results.
- This PR also includes some UI additions to diplay this information nicely.

Further improvements can include :
- Only keeping the most precise CPEs (eliminating CPEs that are a substring of other ones).
- Adding a `cpe:` request type.
- Linking CPEs with CVE databases (for example using [vFeed](https://github.com/toolswatch/vFeed)).

### DIFF
--- a/ivre/xmlnmap.py
+++ b/ivre/xmlnmap.py
@@ -199,6 +199,33 @@ def ignore_script(script):
     return False
 
 
+def cpe2dict(cpe_str):
+    """Helper function to parse CPEs. This is a very partial/simple parser."""
+    # Remove prefix
+    if len(cpe_str) < 5 or cpe_str[:5] != "cpe:/":
+        sys.stderr.write("WARNING, invalid cpe format (%s)\n" % cpe_str)
+        return {"value": cpe_str}
+    cpe_body = cpe_str[5:]
+    parts = cpe_body.split(":", 3)
+    l = len(parts)
+    if l < 2:
+        sys.stderr.write("WARNING, invalid cpe format (%s)\n" % cpe_str)
+        return {"value": cpe_str}
+    cpe_type = parts[0]
+    cpe_vend = parts[1]
+    cpe_prod = parts[2] if l > 2 else ""
+    cpe_comp = parts[3] if l > 3 else ""
+
+    ret = {
+        "type": cpe_type,
+        "vendor": cpe_vend,
+        "product": cpe_prod,
+        "components": cpe_comp,
+        "value": cpe_str,
+    }
+    return ret
+
+
 class NoExtResolver(EntityResolver):
 
     """A simple EntityResolver that will prevent any external
@@ -232,6 +259,12 @@ class NmapHandler(ContentHandler):
         self._curhostnames = None
         self._filehash = filehash
         print "READING %r (%r)" % (fname, self._filehash)
+
+    def _pre_addhost(self):
+        """Executed before _addhost for host object post-treatment"""
+        if 'cpes' in self._curhost:
+            cpes = self._curhost['cpes']
+            self._curhost['cpes'] = cpes.values()
 
     def _addhost(self):
         """Subclasses may store self._curhost here."""
@@ -446,6 +479,9 @@ class NmapHandler(ContentHandler):
                 attrsdict['domains'] = list(
                     utils.get_domains(attrsdict['host']))
             self._curtrace['hops'].append(attrsdict)
+        elif name == 'cpe':
+            # start recording
+            self._curdata = ''
 
     def endElement(self, name):
         if name == 'nmaprun':
@@ -454,6 +490,7 @@ class NmapHandler(ContentHandler):
         elif name == 'host':
             if self._curhost['state'] == 'up' and ('ports' in self._curhost
                                                    or not self._needports):
+                self._pre_addhost()
                 self._addhost()
             self._curhost = None
         elif name == 'hostnames':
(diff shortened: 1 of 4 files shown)


athiasjerome:
in case you would like more https://github.com/athiasjerome/XORCISM/tree/master/Import_all

p-l-:
Great! I'm testing this right now!

ivre/xmlnmap.py
```diff
@@ -199,6 +199,33 @@ def ignore_script(script):
     return False
 
 
+def cpe2dict(cpe_str):
+    """Helper function to parse CPEs. This is a very partial/simple parser."""
+    # Remove prefix
+    if len(cpe_str) < 5 or cpe_str[:5] != "cpe:/":
```
commial: Here, you can use `not cpe_str.startswith` instead of `len(cpe_str) < 5 or cpe_str[:5] != "cpe:/"`

p-l-:
Thanks a lot for this work. Here are some observations:
  - maybe the second column could be smaller to leave room for the next ones (from the examples I have for now it seems to make sense)?
  - usually `clickable` style is used when an element can be clicked (to add a filter). I think it would be nice to add (in another PR) the ability to filter via `cpe` (add a `.searchcpe()` method to `ivre.db.mongo`, add a `cpe:` filter to the WebUI) but while it's not done I'd prefer if the element was not clickable.
  - I have an odd display with a host:

Here is how the result looks like:
cpe

And here is the `cpes` part of the JSON data:
```
    "cpes": [
        {
            "product": "linux_kernel", 
            "vendor": "linux", 
            "origins": [
                "os.osmatch.line:41110", 
                "os.osmatch.line:41350"
            ], 
            "value": "cpe:/o:linux:linux_kernel:2.4", 
            "components": "2.4", 
            "type": "o"
        }, 
        {
            "product": "linux_kernel", 
            "vendor": "linux", 
            "origins": [
                "os.osmatch.line:53103"
            ], 
            "value": "cpe:/o:linux:linux_kernel:2.6", 
            "components": "2.6", 
            "type": "o"
        }, 
        {
            "product": "fritz%21box_fon_wlan_7170", 
            "vendor": "avm", 
            "origins": [
                "os.osmatch.line:53103"
            ], 
            "value": "cpe:/h:avm:fritz%21box_fon_wlan_7170", 
            "components": "", 
            "type": "h"
        }
    ], 
```

ivre/xmlnmap.py
```diff
@@ -199,6 +199,33 @@ def ignore_script(script):
...
+        sys.stderr.write("WARNING, invalid cpe format (%s)\n" % cpe_str)
+        return {"value": cpe_str}
+    cpe_body = cpe_str[5:]
+    parts = cpe_body.split(":", 3)
+    l = len(parts)
```
p-l-: Please use a longest (and more understandable) variable name, such as `nparts` for example.

p-l-:
The odd display seems to occur when the text in the third column is longest than the text in the fourth one.

ivre/xmlnmap.py
fmonjalet: Yes, much, much better. :) Thanks.

fmonjalet:
Hi p-l-,thanks for your remarks, I'm starting to work on it.
Do you have any class for hoverable but not clickable elements? The cpe filter is definitely the next step.

fmonjalet:
I think that I fixed all the remarks, except the odd display. Since the columns are better sized, the bug _appears_ a lot less (did not see it in my samples) but it is still _there_ (can be reproduced with an overly long product name).
It seems to be a rather painful bug for an html/css layman. Would you have any idea to fix that?

By the way, thank you @athiasjerome for the link, I'll take a glance at it.

p-l-:
I've just spent some time trying to figure out a way to fix this odd display, but so far I have nothing better to offer. I'm merging this, but let's keep that in mind in case we find a better idea ;).

BTW, any idea on why the popover do not get the `max-width: 500px;` from `ivre.css` applied?

Anyway, thanks a lot for this work Florent!

PR #32, opened 2015-05-02
WebUI: add display:cpe filter

p-l-:
After PR #31, this adds a filter `display:cpe` to the Web UI to only display the CPE results.

### DIFF
--- a/web/static/templates/view-cpes-only.html
+++ b/web/static/templates/view-cpes-only.html
@@ -1,0 +1,91 @@
+<!--
+  This file is part of IVRE.
+  Copyright 2011 - 2015 Pierre LALET <pierre.lalet@cea.fr>
+
+  IVRE is free software: you can redistribute it and/or modify it
+  under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  IVRE is distributed in the hope that it will be useful, but WITHOUT
+  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+  or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+  License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with IVRE. If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<div class="result-container"
+     ng-repeat="host in results"
+     >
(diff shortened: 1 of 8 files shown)


p-l-:
@fmonjalet please test and merge if that's OK for you.

web/static/templates/view-cpes-only.html
```diff
@@ -0,0 +1,91 @@
...
+
+  IVRE is distributed in the hope that it will be useful, but WITHOUT
+  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+  or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+  License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with IVRE. If not, see <http://www.gnu.org/licenses/>.
+-->
+
```
fmonjalet: This file contains mixed tabs and spaces, maybe you could re-indent with spaces only?

web/static/templates/view-cpes-only.html
```diff
@@ -0,0 +1,91 @@
...
+		ng-repeat="port in summary.ports | orderBy: ['protocol','port']"
+		><!--
+         --><a ng-click="setparam(port.protocol+'/'+port.port)" class="clickable">
+	      {{port.protocol}}/{{port.port}}</a><span ng-if="!$last">, </span>
+	  </span>
+	</td>
+      </tr>
+    </table>
+    <dt class="result-cpe" ng-if="host.cpes">
+      CPE (Common Platform Enumeration)
```
fmonjalet: This remark might be out of the scope of this PR, but it would be nice to have a template include system (with angularjs directives?) to avoid duplicating html code. If an upcoming PR added cpe filtering triggered by clicking on this table's entries, it would be easier to have only one place to modify.

fmonjalet:
Nice feature! Except some superficial remarks, it works well for me.
It seems that js and html files in IVRE tend to contain mixed tabs and spaces, it might be nice to uniformize that (maybe in a separate PR, as it affects other files). What do you think?

p-l-:
Thanks for reporting the mixed tabs & spaces, you're totally right, I'll fix that in another PR.

Template factorization is a more general topic (for example, ports summaries might also be a template alone) that should, also, require another PR.

You can merge this one in the meantime.

PR #33, opened 2015-05-03
WebUI/db: added a CPE filter

fmonjalet:
This follows PR #31 and enables CPE filtering in the web ui. Works very well with display:cpe from PR #32.
@p-l-, does the filter syntax seem reasonable to you?

(*Note: the scancli part has not been done yet.*)

### DIFF
--- a/doc/WEBUI.md
+++ b/doc/WEBUI.md
@@ -202,7 +202,7 @@ single or double quotes.
   - `netdev`, `networkdevice` look for network devices (firewalls,
     routers, ...).
   - `phonedev` look for telephony devices.
-  - `cpe:[value] | cpe:[field=value](, ...)` look for a given cpe, by value or by field (type, vendor, product or version).
+  - `cpe(:[type](:[vendor](:[product](:[version]))))` look for a given cpe. Each field can be a /regex/.
   - `[!]hop:` look for a particular IP address in the traceroute
     results.
   - `[!]hopname:` look for a matching hostname in the traceroute
--- a/doc/WEBUI.md
+++ b/doc/WEBUI.md
@@ -202,7 +202,7 @@ single or double quotes.
   - `netdev`, `networkdevice` look for network devices (firewalls,
     routers, ...).
   - `phonedev` look for telephony devices.
-  - `cpe:[value] | cpe:[field=value](, ...)` look for a given cpe, by value or by field (type, vendor, component or product).
+  - `cpe:[value] | cpe:[field=value](, ...)` look for a given cpe, by value or by field (type, vendor, product or version).
   - `[!]hop:` look for a particular IP address in the traceroute
     results.
   - `[!]hopname:` look for a matching hostname in the traceroute
--- a/doc/WEBUI.md
+++ b/doc/WEBUI.md
@@ -202,6 +202,7 @@ single or double quotes.
   - `netdev`, `networkdevice` look for network devices (firewalls,
     routers, ...).
   - `phonedev` look for telephony devices.
+  - `cpe:` look for a given cpe.
   - `[!]hop:` look for a particular IP address in the traceroute
     results.
   - `[!]hopname:` look for a matching hostname in the traceroute
--- a/doc/WEBUI.md
+++ b/doc/WEBUI.md
@@ -202,7 +202,7 @@ single or double quotes.
   - `netdev`, `networkdevice` look for network devices (firewalls,
     routers, ...).
   - `phonedev` look for telephony devices.
-  - `cpe:` look for a given cpe.
+  - `cpe:[value] | cpe:[field=value](, ...)` look for a given cpe, by value or by field (type, vendor, component or product).
   - `[!]hop:` look for a particular IP address in the traceroute
     results.
   - `[!]hopname:` look for a matching hostname in the traceroute
--- a/ivre/db/mongo.py
+++ b/ivre/db/mongo.py
@@ -1143,30 +1143,30 @@ have no effect if it is not expected)."""
         }}
 
     @staticmethod
-    def searchcpe(value=None, type=None, vendor=None, product=None,
+    def searchcpe(value=None, cpe_type=None, vendor=None, product=None,
                   components=None):
         """Look for a CPE by value (original cpe string), type (a, o or h),
         vendor, product or components (the part after the column following
         the product). No argument will just check for cpe existence.
 
         """
-        if all(arg is None
-               for arg in [value, type, vendor, product, components]):
-            return {"cpes": {"$exists": True}}
-
-        flt = {}
         fields = [
             ("value", value),
-            ("type", type),
+            ("type", cpe_type),
             ("vendor", vendor),
             ("product", product),
             ("components", components),
         ]
-
-        for (field_name, field_val) in fields:
-            if field_val is not None:
-                flt["cpes." + field_name] = field_val
-        return flt
+        flt = dict((field, value) for field, value in fields
+                                  if value is not None)
+        nflt = len(flt)
+        if nflt == 0:
+            return {"cpes": {"$exists": True}}
+        elif nflt == 1:
+            field, value = flt.popitem()
+            return {"cpes.%s" % field: value}
+        else:
+            return {"cpes": {"$elemMatch": flt}}
 
     def topvalues(self, field, flt=None, topnbr=10, sortby=None,
                   limit=None, skip=None, least=False, archive=False,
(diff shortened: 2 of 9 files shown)


doc/WEBUI.md
```diff
@@ -202,6 +202,7 @@ single or double quotes.
   - `netdev`, `networkdevice` look for network devices (firewalls,
     routers, ...).
   - `phonedev` look for telephony devices.
+  - `cpe:[value] | cpe:[field=value](, ...)` look for a given cpe, by value or by field (type, vendor, component or product).
```
p-l-: Don't you think a syntax like `cpe:field:value,field:value` would better fit the syntax used by other filters?

ivre/db/mongo.py
```diff
@@ -1142,6 +1142,32 @@ def searchscreenshot(port=None, protocol='tcp', service=None, neg=False):
                            'screenshot': {'$exists': not neg}}
         }}
 
+    @staticmethod
+    def searchcpe(value=None, type=None, vendor=None, product=None,
```
p-l-: I'd really prefer another name for `type`, such as `cpetype` or `cpe_type` for example (even if it is redundant with the method name `searchcpe`), or maybe `type_` if you like this one better (see [PEP8](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles) `single_trailing_underscore_`).

ivre/db/mongo.py
```diff
@@ -1142,6 +1142,32 @@ def searchscreenshot(port=None, protocol='tcp', service=None, neg=False):
...
+            ("type", type),
+            ("vendor", vendor),
+            ("product", product),
+            ("components", components),
+        ]
+
+        for (field_name, field_val) in fields:
+            if field_val is not None:
+                flt["cpes." + field_name] = field_val
+        return flt
```
p-l-: If I understand your code well, you have a mistake here: if I specify at least two values (say `value=yyy` and `cpe_type=xxx`), the method will return `{"cpes.type": "xxx", "cpes.value: "yyy"}`. This is not what you want, because it would match a document with, for example `{"cpes": [{"type": "xxx", "value": "zzz"}, {"type": "zzz", "value": "yyy"}]}`.

Plus I think this code might simplified a bit. Here is a possibility (not tested):

```Python
    flt = dict(
        (field, value)
        for field, value in [
                ("value", value),
                ("type", cpe_type),
                ("vendor", vendor),
                ("product", product),
                ("components", components),
        ]
        if value is not None
    )
    nfields = len(nfields)
    if nfields == 0:
        return {"cpes": {"$exists": True}}
    if nfields == 1:
        field, value = flt.popitem()
        return {"cpes.%s" % field: value}
    return {"cpes": {"$elemMatch": flt}}
```

What do you think?

web/cgi-bin/scanjson.py
```diff
@@ -460,6 +460,22 @@ def check_referer():
...
+            for cpe_arg in cpe_args:
+                if '=' not in cpe_arg:
+                    # only value
+                    cpe_kwargs["value"] = ivre.utils.str2regexp(cpe_arg)
+                else:
+                    field, value = cpe_arg.split("=", 1)
+                    if field not in allowed_fields:
+                        continue
+                    cpe_kwargs[field] = ivre.utils.str2regexp(value)
+        flt = db.nmap.flt_and(flt, db.nmap.searchcpe(**cpe_kwargs))
```
p-l-: This part can be simplified, like this for example:

```Python
    elif nq == "cpe":
        if q[1]:
            allowed_fields = set(["value", "type", "vendor", "product",
                                  "components"])
            cpe_kwargs = dict(
                (field, value) for field, value in (
                    cpe_arg.split("=", 1)
                    if "=" in cpe_arg else
                    ("value", ivre.utils.str2regexp(cpe_arg))
                    for cpe_arg in q[1].split(',')
                ) if field in allowed_fields
            )
            flt = db.nmap.flt_and(flt, db.nmap.searchcpe(**cpe_kwargs))
        else:
            flt = db.nmap.flt_and(flt, db.nmap.searchcpe())
```

p-l-:
I like this code! Thanks a lot Florent.

Please let me know what you think about my remarks (the most important points being about the WebUI syntax and the `type` argument name).

ivre/db/mongo.py
fmonjalet: You are right (I'm not very familiar with the MongoDB API yet). What would you think of:
```python
        fields = [
            ("value", value),
            ("type", cpe_type),
            ("vendor", vendor),
            ("product", product),
            ("components", components),
        ]
        flt = dict((field, value) for field, value in fields
                                  if value is not None)
        nflt = len(flt)
        if nflt == 0:
            return {"cpes": {"$exists": True}}
        else:
            return {"cpes": {"$elemMatch": flt}}
```
No special case for 1 element match, but shorter code.

p-l-: The problem with using `$elemMatch` with only one element is that it is less efficient in MongoDB, particularly if we index some of the fields.

BTW, maybe you could use this PR to add some indexes to make CPE queries faster?

doc/WEBUI.md
fmonjalet: While this is right, using the same separator between the request type (cpe) and the fields/value seems a little odd to me. Moreover, cpe strings contain multiple ":", and the parser could get messy if we want to keep the "cpe:cpe_string" syntax. That said, if you really prefer the full ":" syntax, I will update the parser accordingly.

What is your preference?

ivre/db/mongo.py
fmonjalet: I've updated my code accordingly (commit to come).

Sure for the indexes, which fields would you index? All the cpe fields?

doc/WEBUI.md
p-l-: You're right of course, let's keep the `=`.

web/cgi-bin/scanjson.py
fmonjalet: While the syntax is more concise, it seems less readable to me. Now that "type" has to be translated to "cpe_type" the code is not really simpler. I can try to find an intermediate solution.
If you prefer this syntax though, this is not a problem to me. :)

ivre/db/mongo.py
p-l-: First of all, I think CPE indexes should be [sparse](http://docs.mongodb.org/manual/core/index-sparse/).

I think at least an index on `value` is mandatory. Then, I'm not sure. We can either opt for
  - an index per field.
  - a [compound index](http://docs.mongodb.org/manual/core/index-compound/) that encompasses all the (other) fields in the correct order (this would be coherent with the choices made for `service`, `product`, and `version`).
  - no other index, and build the `value` to query (or at least a regular expression) based on the provided fields.

In any case, I think we should try, in `searchcpe()`, to complete the query. For example, if the user provides `product="lighttpd"`, we might want to guess `vendor="lighttpd", type="a"` and possibly `value=re.compile("^cpe:/a:lighttpd:lighttpd:")`.

This is particularly useful if we want to go for option 2 or 3 for indexing.

What do you think?

web/cgi-bin/scanjson.py
p-l-: Of course, with the need to translate `cpe_type` / `type`, this will be less readable.

fmonjalet: I came up with:
```python
    elif nq == "cpe":
        if q[1]:
            cpe_kwargs = {}
            allowed_fields = set(["value", "cpe_type", "vendor", "product",
                                  "components"])
            cpe_args = q[1].split(',')
            for cpe_arg in cpe_args:
                field, value = cpe_arg.split("=", 1) if "=" in cpe_arg\
                                                     else ("value", cpe_arg)
                if field == "type":
                    field = "cpe_type"
                if field in allowed_fields:
                    cpe_kwargs[field] = ivre.utils.str2regexp(value)
            flt = db.nmap.flt_and(flt, db.nmap.searchcpe(**cpe_kwargs))
        else:
            flt = db.nmap.flt_and(flt, db.nmap.searchcpe())
```
Do you prefer this verison?

p-l-: Yep. Maybe you can just use `for cpe_arg in q[1].split(','):` since `cpe_args` is not used later.

fmonjalet: You're right, I'm updating it.

doc/WEBUI.md
fmonjalet: Following our discussion, I updated the syntax to be `cpe(:type(:vendor(:product(:version)))` (you can see help.js for more information).

ivre/db/mongo.py
fmonjalet: Following our discussion, `cpes.values` have been removed, and `cpes` are now indexed with a compound index: `{"cpes.type": 1, "cpes.vendor": 1, "cpes.product": 1, "cpes.version": 1}`. What do you think of this solution?

PR #42, opened 2015-05-18
DB/Mongo: compatibility with both pymongo 2 & 3

p-l-:


commial:
You may also update the DOCS accordingly.

PR #45, opened 2015-05-29
JS Refactoring

commial:
This PR divides `ivre.js` in a few submodules, regrouped by feature:
* `ivre/controllers.js`: AngularJS related controllers
* `ivre/graph.js`: Graphing utilities
* `ivre/params.js`: Parameters handling
* `ivre/tooltip.js`: Tooltip handling
* `ivre/utils.js`: Common util functions
* `ivre/ivre.js`: Main IVRE

Currently, these scripts are included in html files thanks to the common `<script>` tag. In the future, may RequireJS or a build pass (like JQuery build process) could be implemented to avoid including all of these scripts.

In addition, as complementary code snippets are regrouped, it is possible to use Javascript namespaces.
There is an example in `tooltip.js`:
* `remove_tooltip` -> `ToolTip.remove`
* `set_tooltip_filter` -> `ToolTip.set_filter`
* ...

If this process is accepted, I'll do the same for at least `params` and `graph` methods.
So please leave a comment before landing the PR (if you're planning to).


p-l-:
Thanks @commial! Great work.

Can you update the `MANIFEST` file (do not believe the first line of that file)?

PR #60, opened 2015-06-30
Update Bootstrap to v3

### LINKED ISSUE #52: Upgrade shipped twitter bootstrap version
Since this is a work in progress, PR including changes to the static part of the Web UI will not be merged until this issue has been closed.

commial:
This PR updates the Bootstrap version to v3 (closes #52).

In addition to common migrations:
* the *Top Values* completion has been rewritten
* the *Top Values* help has been moved from controller to `help.js`, next to the one relative to filters
* resolutions equal and higher than 1024x768 are now supported (see screenshot)

ivre1024

In a future work, methods relative to filter and topvalues may be merged.


p-l-:
@commial, thanks for that (huge) work!

There are some differences between the old interface and the new one. Some are OK, but some seem weird to me. See screenshots below (with Chromium but I get the same with Firefox).

Old:
ivre-webui-old

New:
ivre-webui-new

The top bar is bigger, which I don't really like, but that also has consequences.
  - The "Droids" logo is no longer centered.
  - The "results counter" and the results themselves start to close to the top bar.

The separations in the left bar are missing.

The "note" icons in the upper-right corner of each result are also missing.

You cannot see that in the screenshots, but the buttons below the graphs and maps (save, zoom) have almost disappeared.

The top values completion seem to be broken (type "port:op" for example, or "countports:open").

commial:
Thank you for you vigilance! You're absolutely right.

I've updated the PR. I agree with you, I prefer the small navbar too (may I'm used to it). As it is not the default behavior of Bootstrap v3, it is in a separate commit.
The compatibility with 1024x768 resolution is maintained.

I'll investigate the top values issue.

PR #83, opened 2015-07-12
New feature: Report generation

commial:
This PR adds a report feature, inspired from the [Shodan](https://www.shodan.io/report/oXoaOJc0) one.

This feature aims to provide an easy way to build a summary of the scanning result, according to some filters.
One can personalize the report aspect through a web interface included in the page. This interface can be hide for further exporting / printing.

The report feature is reachable from *Menu > Share > Report*.

Below, a few screen-shots.

Configuration toolbox:
config

Sample report:
report

Responsive on mobile (Nexus 5, Chrome 43):
mobile

Some code snippets have been re-factored to permit their re-use.

Possible ameliorations:
* introduce a size for each elements (1/3, 2/3, 1/2 and 1 page width)
* re-factor elements as module to facilitate further additions
* import / export report template as JSON, to be able to re-use them in several report
* ability to have query-custom elements, different from the one in filters


p-l-:
Thanks @commial! Some minor remarks:

  - the configuration screen renders poorly on "small screen" (tested on 1360x768) devices.
  - the filter does not seem to apply to "top values" but only to "maps"
  - the top nav bar:
    - should have the "home" button at the rightmost position
    - should look exactly like IVRE's top nav bar (droids-corp.org link at the leftmost position)
    - might use "Config" instead of "Configuration" to keep short menu names, even if, for now, we don't need extra room here.

commial:
I'm taking in account your remarks.

Did you mean the "home" button at the *leftmost* (~~right~~) position and droids-corp.org link at the *rightmost* (~~left~~) position (as `/index.html`)?

p-l-:
Yes sorry for the confusion between right and left ;-).

commial:
Does it looks like good for you ?

Another option for configuration display is using the page wide width. But it ends with a very large table in my resolution (1920x1080).

commial:
The layout has been modified to centered the menu, and keep logo, IVRE and link in the corner.
Resolution 1024x768 (laptop / smartphone), 1360x768 and 1920x1080 have been tested.

I've also rebased the PR.

p-l-:
Thanks and congrats for this work!

PR #84, opened 2015-07-12
Reactive web

commial:
This PR aims to speed up the web rendering.

According to [Speeding up AngularJS apps with simple optimizations](https://www.binpress.com/tutorial/speeding-up-angular-js-with-simple-optimizations/135):
```
It’s known that Angular becomes slower with around 2,000 bindings due to the process behind dirty-checking. The less we can add to this limit the better, as bindings can add up without us really noticing it!
```

Using a `$$watchers` [counter](http://blog.yodersolutions.com/angularjs-count-the-number-of-watchers-on-a-page/), on a slow render query (`#1 range:64.170.144.217-69.39.229.61` on ModBus scan):
* Without the PR: 250 359 watchers (!)
* By rendering full view only if `host.fulldisplay`: 25494. On my computer, this is the difference between a "Slow script" box on Firefox and no box.
* By using the AngularJS one-time-binding syntax on summary `{{::`: 8866. Expanding the first result (`67.47.105.226`) go to 29698.
* By using this same syntax on full mode: 18975 with the first result opened.

More optimization could be done. This PR aims to be a first step in order to allow a better render time on big hosts.

### DIFF
--- a/web/static/ivre/ivre.css
+++ b/web/static/ivre/ivre.css
@@ -133,10 +133,6 @@ form {
     margin-bottom: 10px;
 }
 
-.result-hidden-part {
-    display: none;
-}
-
 .summary > table > tbody > tr > td {
     padding-right: 10px;
     white-space: nowrap;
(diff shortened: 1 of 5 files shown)


p-l-:
Really nice catch! There are still a lot bindings that should be switched to one-time bindings in `templates/`. Do you want to fix them in the same PR or do you prefer that I merge this one and do the remainder in another one?

PR #90, opened 2015-07-14
Web: handle tooltip for command with parameters

### LINKED ISSUE #85: WebUI: completion/help broken for some filters
The tooltip disappears for filters with an optional argument when typing ":". Example: "screenwords:".

commial:
Fix #85 .

While typing `screenwor`, two commands are proposed: `screenwords` and `screenwords:`. Is it the wanted behavior?

### DIFF
--- a/web/static/help.js
+++ b/web/static/help.js
@@ -31,6 +31,13 @@ function prepare(help) {
 	    help.content["-" + key] = help.content[key];
 	}
     }
+
+    // Manage optional parameters
+    for (var key in help.content) {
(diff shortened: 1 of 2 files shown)


p-l-:
No it's not. It should display the help with title `screenwords(:<b>[...]`. And this help should remain displayed when typing the following "ds", and when typing ":" (and then when typing anything else, just the same as for filters with mandatory parameters).

commial:
It should now works as expected.

PR #92, opened 2015-07-15
setup.py now uses setuptools instead of distutils

0x90:
setup.py uses setuptools instead of distutils

setup for developers available:
```
python setup.py develop
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#! /usr/bin/env python2
 
 # This file is part of IVRE.
 # Copyright 2011 - 2015 Pierre LALET <pierre.lalet@cea.fr>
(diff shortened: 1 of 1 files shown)


p-l-:
Sorry I merged that PR too soon. Why should we switch to `setuptools`?

0x90:
IMHO, setuptools are the future, and distutils is past. 
Also develop option is available only in the case of setuptools usage.

p-l-:
OK, thanks for taking the time to explain.

I have myself no opinion on that, but I had two "issues" when installing ivre with `setuptools` instead of `distutils` that I'd like to fix prior to switching.

First, the "new" `setup.py` tried to fetch and install pip package for `pymongo` 2.7.2, which is already installed through my Linux distribution's packages. Is there a way to tell `setup.py`. Second, it replaced the scripts in `/usr/loca/bin`with scripts invoking `easy_install`.

After some (quick) research, it seems that the option `--single-version-externally-managed` restores the "old" behaviour. While this option is not the easiest to type or remember, it could be enough. I'll give it another try soon.

Anyway thanks for your time!

PR #104, opened 2015-07-21
WebUI: responsive graphs and logos in reports

p-l-:


commial:
I got a few points.

In my opinion, you should keep the *RATIO* constants, in addition with the `$("#" + chart).width()` code.

Indeed, what you're doing here is the creation of a box with an arbitrary size (for instance, 450 * 500 px). Then, thanks to the `ViewBox` (congrats for it, I completely missed that), the graph elements are resized.

These constants are great for a 1360x768px resolution. But in a bigger resolution, it ends with a graph with very big lines (taking the *Top values* example).

With the PR:
with
Without the PR:
without

I think the normal way should be using the `$...width()` snippet, because it seems to be the usual case. And in the other cases, one could accept the resize (that is to say, big or small lines).

Second point, you should always use a parent `div` with the class `row` when you use the `col-sm-*` template.

Finally, I see you remove the `jumbotron` style from the report header. Is it intentional?
I personally prefer with it, because it marks a difference with the rest of the report. Taste and colors...

p-l-:
Thanks Camille. Is the world map better now?

p-l-:
After discussions w/ @commial, here is an updated version of the PR.

PR #106, opened 2015-07-29
Add an ID filter for the python-api, CLI and WebUI

xavier-martin:


ivre/db/__init__.py
```diff
@@ -271,11 +277,13 @@ def __init__(self):
...
         self.argparser.add_argument('--countports', metavar='COUNT',
-                                    help='show only results with a number of open '
-                                    'ports within the provided range', nargs=2)
+                                    help='show only results with a number '
+                                    'of open ports within the provided range',
+                                    nargs=2)
         self.argparser.add_argument('--no-countports', metavar='COUNT',
-                                    help='show only results with a number of open '
-                                    'ports NOT within the provided range', nargs=2)
+                                    help='show only results with a number '
```
p-l-: Please do not change files when it's not needed, as it makes `git blame` unnecessarily harder to use.

ivre/db/__init__.py
```diff
@@ -260,6 +260,12 @@ def __init__(self):
                                     help='show only results from this source')
         self.argparser.add_argument('--version', metavar="VERSION", type=int)
         self.argparser.add_argument('--timeago', metavar='SECONDS', type=int)
+        self.argparser.add_argument('--id', metavar='ID[,ID[...]]',
```
p-l-: Use `nargs='+'`, which does exactly what you want. This will be `argparse`-specific, so you will need to define a constant `USING_ARGPARSE` (such a constant is already used within some other tools in `bin/`).

Installations lacking `argparse` (that is, those using Python < 2.7 without the external `argparse` module installed) will only be able to select one ID at a time (which is not really a problem, particularly since `optparse` support is to be dropped soon).

ivre/db/mongo.py
```diff
@@ -1127,6 +1127,19 @@ def searchports(self, ports, protocol='tcp', state='open', neg=False):
             for port in ports]}}
 
     @staticmethod
+    def searchobjectid(listId=None, neg=False):
+        assert listId is not None
```
p-l-: Is there a reason I have missed why you do not want to use a mandatory parameter here?

IMO, the method "prototype" should be:
```Python
@staticmethod
def searchobjectid(oid, neg=False):
 ```

Then, you should consider that either a single object ID has been passed, or a list (or tuple, or even a generator, maybe from a MongoDB cursor for example).

By the way, when adding new classes or methods, please make sure to add an appropriate docstring with it.

Since `bson.objectid.ObjectId()` constructor will happily handle an argument that happens to be already an `ObjectId`, it is OK to call it for each element of the list regardless of its type, but the docstring should make it clear that it is OK to use either strings or object IDs.

File: ivre/db/mongo.py [hunk 4]
```diff
@@ -1143,7 +1156,7 @@ def searchcountopenports(minn=None, maxn=None, neg=False):
             return {'$or': [{'openports.count': cond} for cond in flt]}
         # return {'openports.count':
         #         dict(item for cond in flt for item in cond.iteritems())}
-        return {'openports.count': {'$lte':maxn, '$gte':minn}}
+        return {'openports.count': {'$lte': maxn, '$gte': minn}}
```
p-l-: This change has nothing to do here as it only breaks the git history.

ivre/db/mongo.py
```diff
@@ -2028,6 +2041,12 @@ def parse_args(self, args, flt=None):
             flt = self.flt_and(flt, self.searchsource(args.source))
         if args.timeago is not None:
             flt = self.flt_and(flt, self.searchtimeago(args.timeago))
+        if args.id is not None:
```
p-l-: You will have to update that also: because of the `nargs="+"`, `args.id` will be a list when using `argparse`, and a single value when using `optparse`.

ivre/db/mongo.py
```diff
@@ -2073,10 +2092,13 @@ def parse_args(self, args, flt=None):
             flt = self.flt_and(flt, self.searchopenport(neg=True))
         if args.countports:
             minn, maxn = int(args.countports[0]), int(args.countports[1])
-            flt = self.flt_and(flt,self.searchcountopenports(minn=minn, maxn=maxn))
+            flt = self.flt_and(flt, self.searchcountopenports(minn=minn,
```
p-l-: Same here, do not change files when it is not needed.

ivre/webutils.py
```diff
@@ -245,6 +245,10 @@ def add_unused(neg, param, value):
                 limit = min(limit, config.WEB_MAXRESULTS)
         elif param == "archives":
             archive = not neg
+        elif param == "id":
+            list_id = [str(oid) for oid in value.replace('-', ',').split(',')]
```
p-l-: Why do you need the `str()` constructor here? Wouldn't `oids = value.replace('-', ',').split(',')` be enough?

And if so, defining a variable to use it once makes little sense, write:

```Python
    flt = db.nmap.flt_and(flt, db.nmap.searchobjectid(
        value.replace('-', ',').split(','), neg=neg))
```

web/static/help.js
```diff
@@ -76,6 +76,10 @@ var HELP_FILTERS = {
 	    "title": "<b>(!)</b>archives",
 	    "content": "Look for archived results. <code>!archives</code> has no effect, since it is the default behavior.",
 	},
+	"id:": {
+	    "title": "<b>(!)</b>id:<b>[ObjectID],[ObjectID],..</b>",
```
p-l-: `"title": "<b>(!)</b>id:<b>[object ID](,[object ID](,...))</b>",`

ivre/db/mongo.py
xavier-martin: It seems that `args.id` can still be `None` even with the use of `nargs="+"`, should I write
`if args.id:` instead ?

p-l-: No, I think the correct test is still `args.id is not None`.

ivre/db/__init__.py
```diff
@@ -242,8 +242,10 @@ def __init__(self):
         try:
             import argparse
             self.argparser = argparse.ArgumentParser(add_help=False)
+            USING_ARGPARSE = True
         except ImportError:
-            self.argparser = utils.FakeArgparserParent()
```
p-l-: 1. Is there a reason why you no longer use the wrapper `utils.FakeArgparserParent()`
2. You do not define `USING_ARGPARSE = False`, which will cause the script to crash at line 265.

ivre/db/mongo.py
```diff
@@ -1127,6 +1127,20 @@ def searchports(self, ports, protocol='tcp', state='open', neg=False):
             for port in ports]}}
 
     @staticmethod
+    def searchobjectid(oid, neg=False):
+        """Filters records by their ObjectID.
+        oid can either be a single ObjectID in string or bson.objectid.ObjectID
```
p-l-: I'd prefer something like:
```
`oid` can be a single or many (as a list or any iterable) object ID(s), specified as strings or an `ObjectID`s.
```

ivre/db/mongo.py
```diff
@@ -1127,6 +1127,20 @@ def searchports(self, ports, protocol='tcp', state='open', neg=False):
...
+        format, or an iterable of ObjectIDs"""
+        if not hasattr(oid, '__iter__'):
```
p-l-: I think a test like `isinstance(oid, basestring) or isinstance(oid, bson.objectid.ObjectId)` is more appropriate.

Also, you could handle, as you did in the previous version, iterables with only one item.

For that, you could make sure oid is a list before doing the work. For example:
```
if isinstance(oid, basestring) or isinstance(oid, bson.objectid.ObjectId):
    oid = [bson.objectid.ObjectId(oid)]
else:
    oid = [bson.objectid.ObjectId(elt) for elt in oid]
```

After that, `oid` is always a list of `ObjectId`s.

ivre/db/mongo.py
```diff
@@ -1127,6 +1127,20 @@ def searchports(self, ports, protocol='tcp', state='open', neg=False):
...
+            return {'_id': {'$ne': oid} if neg else oid}
```
p-l-: You do not make sure that `oid` is an `ObjectId` (it could be a string).

ivre/db/__init__.py
xavier-martin: 1. Not really, I clumsily took that from bin/scancli
2. Right, fixed

ivre/db/mongo.py
xavier-martin: Noted, thanks for your vacation time ;)

ivre/db/mongo.py
```diff
@@ -1127,6 +1127,24 @@ def searchports(self, ports, protocol='tcp', state='open', neg=False):
             for port in ports]}}
 
     @staticmethod
+    def searchobjectid(oid, neg=False):
+        """Filters records by their ObjectID.
+        `oid` can be a single or many (as a list or any iterable) object ID(s),
+        specified as strings or an `ObjectID`s."""
+        if isinstance(oid, basestring) or isinstance(oid, bson.objectid.ObjectId):
```
p-l-: You are going to hate me but I had a doubt and checked: there is a shorter (and hence better, as @commial usually says) way to write the test: `if isinstance(oid, (basestring, bson.objectid.ObjectId)):`.

xavier-martin: I do agree it's better, not sure we can still say 'shorter' though.. !

ivre/db/mongo.py
```diff
@@ -1127,6 +1127,24 @@ def searchports(self, ports, protocol='tcp', state='open', neg=False):
...
+            oid = [bson.objectid.ObjectId(oid)]
+        else:
+            oid = [bson.objectid.ObjectId(elt) for elt in oid]
+        if len(oid) == 1:
+            return {'_id': {'$ne': oid[0]} if neg else oid[0]}
+        if neg:
+            return {'$and': [{'_id': {'$ne': bson.objectid.ObjectId(elt)}}
```
p-l-: Avoid, when possible, `$and` and `$or` operators since they make it harder for MongoDB to use indexes. Plus when possible it often makes easier-to-read code.

This part of the code (the `if neg:` and the `else:` blocks) should be rewritten as:
```Python
return {'_id': {'$nin' if neg else '$in': oid}}
```

ivre/db/mongo.py
p-l-: As @commial also says, "if it's better, then it's probably shorter".

ivre/db/mongo.py
xavier-martin: Ok, it's also shorter :)

PR #120, opened 2015-08-17
Fix Unix file name parsing in ftp-anon output

### LINKED ISSUE #118: Wrong filenames in ftp-anon structured data
When symlinks are found, the `filename` value of the file entry is `"filename": [filename] -> [linktaret]`. It should be `"filename": [filename], "linktarget": [linktarget]`.

p-l-:
Fixes #118

### DIFF
--- a/ivre/xmlnmap.py
+++ b/ivre/xmlnmap.py
@@ -264,13 +264,13 @@ def add_ftp_anon_data(script):
         "%(month)s\\s+%(day)s\\s+(?:%(year)s|%(time)s)" % subexprs,
         "%(day)s\\.\\s+%(month)s\\s+%(time)s" % subexprs,
     ])
-    subexprs["fname"] = "%(fname)s(?:\\ \\-\\>\\ %(fname)s)?" % subexprs
     exprs = re.compile("^(?:" + "|".join([
         # unix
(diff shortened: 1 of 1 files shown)


p-l-:
For those who had already migrated to document schema version 3 before this PR has been merged (considering the delay between the two merges, I'd say this is unlikely, but let's keep the script for reference here), here is a Python script to update `ftp-anon` results:

```Python
import re

from ivre.db import db

LINK = re.compile('^(?P<filename>.*)' + re.escape(" -> ")
                  + '(?P<linktarget>.*)$', re.MULTILINE)

for host in db.nmap.get(db.nmap.searchfile(LINK, scripts=['ftp-anon'])):
    update = False
    for port in host.get('ports', []):
            if script['id'] == 'ftp-anon':
                for fileentry in script['ls'].get(
                        'volumes', {'files': []})[0].get(
                            'files', []):
                    link = LINK.match(fileentry['filename'])
                    if link is not None:
                        fileentry.update(link.groupdict())
                        print fileentry
                        update = True
    if update:
        db.nmap.db[db.nmap.colname_hosts].update(
            {'_id': host['_id']},
            {'$set': {
                'ports': host['ports'],
            }},
        )
```

PR #143, opened 2015-09-01
Analyser integration

xavier-martin:
Implement a clustering and anomaly-detection CLI tool.

### DIFF
--- a/bin/analysercli
+++ b/bin/analysercli
@@ -1,0 +1,257 @@
+#! /usr/bin/env python
+#
+# This file is part of IVRE.
+# Copyright 2011 - 2015 Pierre LALET <pierre.lalet@cea.fr>
+#
+# IVRE is free software: you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# IVRE is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+# License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with IVRE. If not, see <http://www.gnu.org/licenses/>.
+
+"""Command line interface tool to access the analyser API"""
+
+from ivre import webutils
+from ivre.db import db
+from ivre.cliutils import colorize_log, StackCli
+from ivre.analyser import Context
+
+
+class AnalyserCli(StackCli):
+    """
+    Command line utility wrapper to access analyser API
+    """
+    def __init__(self):
+        StackCli.__init__(self)
+        self._dfilter = {}
+
+    def preloop(self):
+        context = Context()
+        colorize_log('info', 'Initialization')
+        self._build_filter(raw_input('Filter > '))
+        colorize_log('info', 'Reading database..')
+        context.dfilter = self._dfilter
+        context.init_hosts()
+        self.push(context)
+
+    def postloop(self):
+        colorize_log('info', 'Leaving..')
+
+    @staticmethod
+    def _flt_param(param):
+        """
+        Return a tuple (neg, param), where `neg` is True if the param was
+        negative (i.e. starts with '!') and `param` is the argument stripped
+        from the potential heading '!'
+        """
+        return (True, param[1:]) if param.startswith('!') else (False, param)
+
+    def _build_filter(self, cmd):
+        """Build a db.nmap.flt from the string `cmd`"""
+        query = []
+        cmd = cmd.split()
+        while cmd:
+            param = cmd.pop(0)
+            neg, param = self._flt_param(param)
+            value = cmd.pop(0)
+            if param == 'limit':
+                self._dfilter['limit'] = int(value)
+            else:
+                query.append((neg, param, value))
+        (self._dfilter['filter'], _, _, _,
+         self._dfilter['skip'], _) = webutils.flt_from_query(
+             query, base_flt=db.nmap.flt_empty)
+
+    def do_count(self, _):
+        """Print host count to stdout"""
+        colorize_log('info', "Number of hosts:\n\t%s" % '\n\t'.join(
+            '%s:\t%s' % (key, value)
+            for key, value in self.get_context().count().iteritems()
+        ))
+
+    def do_plot2d(self, _):
+        """Plot hosts in 2 dimensions"""
+        self.get_context().plot(plot3d=False)
+
+    def do_plot3d(self, _):
+        """Plot hosts in 3 dimensions"""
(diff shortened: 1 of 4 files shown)


bin/analysercli
```diff
@@ -0,0 +1,283 @@
...
+from ivre.cliutils import COLORS, StackCli
+from ivre.analyser import Context
+
+
+# --------------------------------------------------------------------------- #
+
+
+class AnalyserCli(StackCli):
+    """
+    Base class for stacked contexts CLI
```
p-l-: How is it different from its parent?

bin/analysercli
```diff
@@ -0,0 +1,276 @@
...
+
+
+class AnalyserCli(StackCli):
+    """
+    Command line utility wrapper to access analyser API
+    """
+    # COLORIZE levels
+    LOG_USAGE = 0
+    LOG_INFO = 10
+    LOG_WARNING = 20
```
p-l-: Maybe you could use a `dict` like:
```
LOG_LEVELS = {
    LOG_USAGE: {"name": "USAGE", "color": "green"},
    [...]
}
```

That could make the code in `_colorized_log()` better.

Also, what about moving that to the `cliutils` sub-module?

xavier-martin: That's a good idea !

ivre/analyser.py
```diff
@@ -0,0 +1,667 @@
...
+        """
+        min_size = []
+        max_size = []
+        dim = hosts.shape[1]
+        for idx in xrange(dim):
+            min_size.append(round(hosts[:, idx].min() - 1))
+            max_size.append(round(hosts[:, idx].max() + 1))
+        return (min_size, max_size)
+
+    def plot(self, dim=2):
```
p-l-: If the only valid values for `dim` are 2 and 3, use a boolean like `plot3d` with default value `False`.

bin/analysercli
```diff
@@ -0,0 +1,263 @@
+#! /usr/bin/env python
+# -*- coding:utf-8 -*-
```
p-l-: Why is this needed?

xavier-martin: Is it not a good practice to always start the code with this statement ?

p-l-: If all your code is plain ASCII, I don't understand why you would need that.

bin/analysercli
```diff
@@ -0,0 +1,262 @@
...
+        context.dfilter = self._dfilter
+        context.init_hosts()
+        self.push(context)
+
+    def postloop(self):
+        colorize_log('info', 'Leaving..')
+
+    @staticmethod
+    def flt_param(param):
+        """Search for parameter starting with '!'"""
```
p-l-: 1. The function name and the docstring are not really consistent with what the function actually does.
2. If this function is only used from a "private" method, it should probably also be "private", or have I missed something?
3. Please avoid unneeded variables:
```Python
if param.startswith('!'):
    return True, param[1:]
return False, param
```

Or better:
```Python
return (True, param[1:]) if param.startswith('!') else (False, param)
```

xavier-martin: You're right, it's shorter thus better, thanks !

bin/analysercli
```diff
@@ -0,0 +1,274 @@
...
+        """Help for do_count command"""
+        log = 'count\n'
+        log += '\tPrint host count to sdout'
+        colorize_log('usage', log)
+
+    @staticmethod
+    def help_plot2d():
+        """Print help for do_plot2d command"""
+        log = 'plot2d\n'
+        log += '\tPlot hosts in two dimensions'
```
p-l-: This is bad. You need to avoid `+` on strings (particularly when they're useless). Here write:
```Python
log = """plot2d                                                                 
\tPlot hosts in two dimensions"""
```
or:
```Python
log = "plot2d\n\tPlot hosts in two dimensions"
```
(both are fine for me)

bin/analysercli
```diff
@@ -0,0 +1,274 @@
...
+# (at your option) any later version.
+#
+# IVRE is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+# License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with IVRE. If not, see <http://www.gnu.org/licenses/>.
+
```
p-l-: Please add a module docstring.

bin/analysercli
```diff
@@ -0,0 +1,274 @@
...
+                self._dfilter['limit'] = int(value)
+            else:
+                query.append((neg, param, value))
+        (self._dfilter['filter'], _, _, _,
+         self._dfilter['skip'], _) = webutils.flt_from_query(
+             query, base_flt=db.nmap.flt_empty)
+
+    def do_count(self, _):
+        """Print host count to stdout"""
+        log = 'Number of hosts:\n'
```
p-l-: Here again to avoid `+` on strings:
```Python
colorize_log('info', "Number of hosts:\n%s" % ''.join(
    '\t%s:\t%s\n' % (key, value)
    for key, value in self.get_context().count().iteritems()
))
```
or (better?):
```Python
colorize_log('info', "Number of hosts:\n\t%s\n" % '\n\t'.join(
    '%s:\t%s' % (key, value)
    for key, value in self.get_context().count().iteritems()
))
```

bin/analysercli
xavier-martin: Ok, noted !

bin/analysercli
xavier-martin: Done.

ivre/analyser.py
```diff
@@ -0,0 +1,675 @@
...
+        """
+        Return a new `Context` which only contains cluster centers
+        Depending on self._clusters type, centers might be one host
+        per cluster or every core-cluster hosts
+        """
+        if self._clusters is None:
+            raise RuntimeError('self._clusters must be initialised first')
+        new_list_id, new_matrix = self._clusters.select_centers(self._list_id,
+                                                                self._matrix)
+        cmd_name = 'sel-centers-' + str(self._clusters)
```
p-l-: Please use `'sel-centers-%s' % self._clusters` instead.

ivre/analyser.py
```diff
@@ -0,0 +1,675 @@
...
+        """
+        Plot n_comp most relevant components for the first dimensions
+        """
+        assert dimension <= 3
+        # Prevent from index error in case PCA went on fewer dimension
+        n_comp = min(n_comp, len(pca_indexes[0]), len(pca_indexes[1]))
+        fig = plt.figure()
+        AXIS = ['X', 'Y', 'Z']
+        for dim in xrange(dimension):
+            plot = fig.add_subplot(dimension, 1, dim+1)
```
p-l-: PEP8: `dim + 1`

p-l-:
Thanks for your work @xavier-martin!

PR #144, opened 2015-09-01
WebUtils: Make flt_from_query context independant

xavier-martin:


ivre/webutils.py
```diff
@@ -244,7 +244,7 @@ def get_init_flt():
     return DEFAULT_INIT_QUERY
 
 
-def flt_from_query(query):
+def flt_from_query(query, flt=db.nmap.flt_empty):
```
p-l-: I think the name of the parameter should be `base_flt` for example, and its default value `None`.

In the code, you could:
```
    flt = get_init_flt() if base_flt is None else base_flt
```

What do you think?

xavier-martin: Right, I think it's a good idea !

PR #151, opened 2015-09-04
HTML export

p-l-:
This includes and hence fixes #150.

### DIFF
--- a/web/static/ivre/content.js
+++ b/web/static/ivre/content.js
@@ -660,10 +660,14 @@ prepare(HELP_TOPVALUES);
 
 /************* Menu content ****************/
 
-/* Menus structure is as follow:
+/* Menus structure is as follows:
  *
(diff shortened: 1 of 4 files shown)


commial:
Instead of `ng_init`, maybe we can use the same mechanisms than for `share` option (in `MENU`).

Aside this remark, I completely agree with the rest of the patch

p-l-:
@commial agreed. Is it OK now?

commial:
Yeah, looks good for me!
May you can just add two comment lines describing the option, as for `share`: 
`* - share: (optional) if set to true, add a 'Share' menu. Default is false`

This way, you can omit `false` value, as they are described and already handled thanks to `ng-if`.
 
Otherwise, I'm OK for the merge.

p-l-:
OK, PR updated.

PR #153, opened 2015-09-10
Label implementation

xavier-martin:


ivre/db/mongo.py
```diff
@@ -721,6 +721,52 @@ def getscreenshot(port):
         if url == "field":
             return port.get('screendata')
 
+    def set_label(self, flt, group, label, archives=False):
+        """Adds `label` for the session `group` to every host matching `flt`"""
```
p-l-: "session"? => """Adds `label` in `group` to every host matching `flt`"""

ivre/db/mongo.py
```diff
@@ -721,6 +721,52 @@ def getscreenshot(port):
...
+        colname = self.colname_oldhosts if archives else self.colname_hosts
+        for host in self.find(colname, flt):
+            labels = host.setdefault('labels', [])
+            try:
+                g_label = [lab for lab in labels if lab['group'] == group][0]
```
p-l-: Do you really need to process the whole `labels` list when the first one might be the one you're looking for? If you want to keep a comprehension construction (and you know I won't blame you for that), use:
```Python
g_label = (lab for lab in labels if lab['group'] == group).next()
```
Obviously the exception you want to catch after that will be `StopIteration`.

ivre/db/mongo.py
```diff
@@ -721,6 +721,52 @@ def getscreenshot(port):
         if url == "field":
             return port.get('screendata')
 
+    def set_label(self, flt, group, label, archives=False):
```
p-l-: I don't really like this function, mainly because you will run an `.update()` even when the record has not been changed.

ivre/db/mongo.py
```diff
@@ -1164,6 +1210,19 @@ def searchcategory(cat, neg=False):
         return {'categories': cat}
 
     @staticmethod
+    def searchlabel(group, label=None, neg=False):
```
p-l-: It would be nice if both `group` and `label` could be regexp.

Also does `neg` work both when there is a matching `group` without a matching `label` and when there is no matching `group`?

ivre/db/mongo.py
xavier-martin: I did not know this way of building generators, I have learned something !

ivre/db/mongo.py
xavier-martin: No it did not, it does by now.

ivre/db/mongo.py
xavier-martin: Ok, it should be fixed now.

ivre/db/mongo.py
```diff
@@ -721,6 +721,52 @@ def getscreenshot(port):
         if url == "field":
             return port.get('screendata')
 
+    def set_label(self, flt, group, label, archives=False):
+        """Adds `label` in `group` to every host matching `flt`"""
+        colname = self.colname_oldhosts if archives else self.colname_hosts
+        for host in self.find(colname, flt):
+            labels = host.setdefault('labels', [])
+            try:
+                g_label = (lab for lab in labels if lab['group'] == group).next()
```
p-l-: You should only catch the exception for this line (and use an `else:` block for the remaining code).

ivre/db/mongo.py
```diff
@@ -721,6 +721,52 @@ def getscreenshot(port):
...
+                if label in g_label['tags']:
+                    continue
+                g_label['tags'].append(label)
+            except StopIteration:
+                # group did not exist.
+                g_label = {'group': group, 'tags': [label]}
+                labels.append(g_label)
+            self.db[colname].update_one({"_id": host['_id']},
+                                        {"$set": {'labels': host['labels']}})
```
p-l-: `{"$set": {'labels': labels}}` should be enough.

ivre/db/mongo.py
```diff
@@ -721,6 +721,52 @@ def getscreenshot(port):
...
+        """
+        colname = self.colname_oldhosts if archives else self.colname_hosts
+        flt = self.flt_and(flt, self.searchlabel(group, label=label))
+        for host in self.find(colname, flt):
+            if label:
+                g_label = [lab for lab in host['labels']
+                           if lab['group'] == group][0]
+                g_label['tags'].remove(label)
+            if not label or not g_label['tags']:
+                host['labels'] = [lab for lab in host['labels']
```
p-l-: The remark about generator vs list also applies here.

ivre/db/mongo.py
```diff
@@ -721,6 +721,52 @@ def getscreenshot(port):
...
+                    {"_id": host['_id']}, {"$set": {'labels': host['labels']}})
+            else:
+                self.db[colname].update_one({"_id": host['_id']},
+                                            {"$unset": {'labels': ''}})
+
+    def clear_labels(self, flt, archives=False):
+        """Removes every labels from hosts matching `flt'"""
+        self.db[
+            self.colname_oldhosts if archives else self.colname_hosts
+        ].update_many(flt, {'$unset': {'labels': ''}})
```
p-l-: Please replace `''` with `True`.

ivre/db/mongo.py
```diff
@@ -1163,6 +1209,27 @@ def searchcategory(cat, neg=False):
                 return {'categories': {'$in': cat}}
         return {'categories': cat}
 
+    def searchlabel(self, group, label=None, regexp=False, neg=False):
```
p-l-: This is not the expected API. I should provide either a regular expression, or a string, and the method should figure out what to do with it. See other methods, such as `.searchcategory()`.

ivre/db/mongo.py
```diff
@@ -2236,6 +2303,20 @@ def parse_args(self, args, flt=None):
             flt = self.flt_and(
                 flt,
                 self.searchservicescript(utils.str2regexp(args.service)))
+        if args.label is not None:
```
p-l-: Maybe you could use `utils.str2regexp`? (when `.searchlabel` has been fixed).

ivre/webutils.py
```diff
@@ -298,6 +298,10 @@ def add_unused(neg, param, value):
             flt = db.nmap.flt_and(flt, db.nmap.searchrange(
                 *value.replace('-', ',').split(',', 1),
                 neg=neg))
+        elif param == "label":
```
p-l-: Maybe you could use `utils.str2regexp`? (when `.searchlabel()` has been fixed).

ivre/db/mongo.py
xavier-martin: Sure, I have updated the code according to your comments

ivre/db/mongo.py
```diff
@@ -721,6 +721,53 @@ def getscreenshot(port):
...
+                g_label = (lab for lab in labels if lab['group'] == group).next()
+            except StopIteration:
+                # group did not exist.
+                g_label = {'group': group, 'tags': [label]}
+                labels.append(g_label)
+            else:
+                if label in g_label['tags']:
+                    continue
+                g_label['tags'].append(label)
+            self.db[colname].update_one({"_id": host['_id']},
```
p-l-: Sorry again, missed that in my previous reviews, `.update_one()` and `update_many()` only exist in PyMongo >= 3.0.

ivre/db/mongo.py
```diff
@@ -721,6 +721,53 @@ def getscreenshot(port):
...
+                                  if lab['group'] != group]
+            if host['labels']:
+                self.db[colname].update_one(
+                    {"_id": host['_id']}, {"$set": {'labels': host['labels']}})
+            else:
+                self.db[colname].update_one({"_id": host['_id']},
+                                            {"$unset": {'labels': ''}})
+
+    def clear_labels(self, flt, archives=False):
+        """Removes every labels from hosts matching `flt'"""
```
p-l-: every label*

ivre/db/mongo.py
```diff
@@ -721,6 +721,53 @@ def getscreenshot(port):
         if url == "field":
             return port.get('screendata')
 
+    def set_label(self, flt, group, label, archives=False):
+        """Adds `label` in `group` to every host matching `flt`"""
+        colname = self.colname_oldhosts if archives else self.colname_hosts
+        for host in self.find(colname, flt):
```
p-l-: Please use `self.get()` instead of `self.find()`.

ivre/db/mongo.py
```diff
@@ -721,6 +721,53 @@ def getscreenshot(port):
...
+
+
+    def remove_label(self, flt, group, label=None, archives=False):
+        """Removes `label` of session `group` from every host matching `flt`
+        If label is None, removes the entire group.
+
+        """
+        colname = self.colname_oldhosts if archives else self.colname_hosts
+        flt = self.flt_and(flt, self.searchlabel(group, label=label))
+        for host in self.find(colname, flt):
```
p-l-: Please use `self.get()` instead of `self.find()`.

ivre/db/mongo.py
```diff
@@ -721,6 +721,53 @@ def getscreenshot(port):
...
+                g_label['tags'].remove(label)
+            if not label or not g_label['tags']:
+                host['labels'] = [lab for lab in host['labels']
+                                  if lab['group'] != group]
+            if host['labels']:
+                self.db[colname].update_one(
+                    {"_id": host['_id']}, {"$set": {'labels': host['labels']}})
+            else:
+                self.db[colname].update_one({"_id": host['_id']},
+                                            {"$unset": {'labels': ''}})
```
p-l-: Please replace `""` by `True`.

ivre/db/mongo.py
```diff
@@ -721,6 +721,53 @@ def getscreenshot(port):
...
+    def remove_label(self, flt, group, label=None, archives=False):
+        """Removes `label` of session `group` from every host matching `flt`
+        If label is None, removes the entire group.
+
+        """
+        colname = self.colname_oldhosts if archives else self.colname_hosts
+        flt = self.flt_and(flt, self.searchlabel(group, label=label))
+        for host in self.find(colname, flt):
+            if label:
+                g_label = (lab for lab in host['labels']
```
p-l-: You're using `host['labels']` a lot here. Maybe you could use a `labels` variable for that?

ivre/db/mongo.py
```diff
@@ -721,6 +721,53 @@ def getscreenshot(port):
         if url == "field":
             return port.get('screendata')
 
+    def set_label(self, flt, group, label, archives=False):
+        """Adds `label` in `group` to every host matching `flt`"""
+        colname = self.colname_oldhosts if archives else self.colname_hosts
+        for host in self.find(colname, flt):
+            labels = host.setdefault('labels', [])
```
p-l-: Since you do not use `host['labels']` after that, you can replace `.setdefault()` by `.get()`.

ivre/db/mongo.py
```diff
@@ -721,6 +721,55 @@ def getscreenshot(port):
         if url == "field":
             return port.get('screendata')
 
+    def set_label(self, flt, group, label, archives=False):
```
p-l-: Sorry I missed that, please use `archive` and not `archives` (I know that's already inconsistent in the current code, but that should be fixed in another PR).

ivre/db/mongo.py
```diff
@@ -721,6 +721,55 @@ def getscreenshot(port):
...
+        """Adds `label` in `group` to every host matching `flt`"""
+        colname = self.colname_oldhosts if archives else self.colname_hosts
+        for host in self.get(flt, archives):
```
p-l-: Please specify the optional parameter name: `self.get(flt, archive=archive)`.

PR #154, opened 2015-09-10
Add tagging features to analyzercli and scancli

xavier-martin:


ivre/analyser.py
```diff
@@ -0,0 +1,726 @@
...
+        new_matrix = matrix[self._center_hosts_idx]
+        return new_list_id, new_matrix
+
+    def plot(self, plot, hosts, plot3d=False):
+        """
+        Plot hosts by clusters and their closest cluster-center host.
+        hosts must be a numpy matrix with 2 or 3 columns
+        depending on plot3d is False or True
+        """
+        DIM = 3 if plot3d else 2
```
p-l-: Please use lowercase for local variable names.

ivre/analyser.py
```diff
@@ -0,0 +1,726 @@
...
+        """
+        Return a new list_id and numpy matrix
+        which only contains hosts from the clusters
+        labeled in `cluster_labels` list
+        """
+        cluster_mask = np.zeros(len(list_id), dtype=bool)
+        for lab in cluster_labels:
+            cluster_mask = np.logical_or((self._cluster_labels == lab),
+                                         cluster_mask)
+        indicies_to_keep = [i for i, keep in enumerate(cluster_mask) if keep]
```
p-l-: By `indicies`, do you mean `indexes`???

xavier-martin: aouch.. right !

p-l-:
@xavier-martin can you please update the PR so that I can use the GitHub interface to review it? (for now it includes changes already merged with #143)

bin/scancli
```diff
@@ -476,6 +476,18 @@ def displayhost_csv(fields, separator, nastr, dic, out=sys.stdout):
                         for line in utils.doc2csv(dic, fields, nastr=nastr)))
     out.write('\n')
 
+def _init_context(hostfilter, skip=None, limit=None):
+    """Return an ivre.analyser.Context initialized object"""
+    from ivre.analyser import Context
```
p-l-: Please move that `import` to the beginning of the file.

bin/scancli
```diff
@@ -476,6 +476,18 @@ def displayhost_csv(fields, separator, nastr, dic, out=sys.stdout):
                         for line in utils.doc2csv(dic, fields, nastr=nastr)))
     out.write('\n')
 
+def _init_context(hostfilter, skip=None, limit=None):
```
p-l-: Since this is not a module, maybe it's not necessary to use the private function name convention here? But I'd like the name to be more explicit. Maybe `analyser_init_context`?

ivre/analyser.py
```diff
@@ -327,6 +347,34 @@ def count(self):
...
+        """
+        Tag hosts in database with their respective cluster number
+        and anomaly status.
+        `session` must be None or a positive integer representing
+        the group tag in database. The possible group tags are
+        'analyse', 'analyse_1', 'analyse_2', ...
+        """
+        session_name = 'analyse'
+        if session and int(session) > 0:
+            session_name += '_' + str(int(session))
```
p-l-: When possible, prefer string formatting (`session_name += '_%d' + int(session)`).

Here, you can even drop the `int(session) > 0` test (since the docstring is very clear about what is expected here) and write:
```Python
session_name = 'analyse' if session is None else 'analyse_%d' % int(session) 
```

ivre/analyser.py
```diff
@@ -327,6 +347,34 @@ def count(self):
...
+    def clean_labels(self, session=None):
+        """
+        Remove tag made by the analyser module (i.e. from group "analyse[_n]")
+        from the database.
+        If n=`session` is provided as a positive integer, the group "analyse[_n]"
+        is cleared instead of "analyse"
+        """
+        session_name = 'analyse'
+        if session and int(session) > 0:
+            session_name += '_' + str(int(session))
```
p-l-: Same remark here, obviously.

ivre/analyser.py
```diff
@@ -579,9 +627,12 @@ def search_anomalies(self, outliers, percent_mode=False):
         instead of `outliers` anomalies
         Update self._anomalies attribute
         """
-        # 0.01 is faster and works well most of the time
-        # but 0.5 works with small host number.
-        nu_percent = outliers * 0.01 if percent_mode else 0.5
+        if percent_mode:
+            nu_percent = outliers * 0.01
+        else:
+            # 0.01 is faster but cause troubles with small host number
```
p-l-: Typo: cause**s**.
Maybe **a** small host number (same for next line).

p-l-:
Thanks, that's much easier to read!

bin/scancli
xavier-martin: Sure, I have changed it !

ivre/analyser.py
xavier-martin: Ok, I have to think about string formatting from now on.

p-l-:
The tests currently fail because you need to add `matplotlib` to `requirements.txt`. If you use recent APIs, you can specify a minimum version to install.

This new dependency needs to be added to (at least) `setup.py`, `buildrpm`, `docker/base/Dockerfile`, `docker/base-local/Dockerfile`

Also, you might want to update the documentation files (at least `FAST-INSTALL-AND-FIRST-RUN`, `INSTALL`, `README`).

bin/scancli
```diff
@@ -17,6 +17,7 @@
 # along with IVRE. If not, see <http://www.gnu.org/licenses/>.
 
 from ivre import utils, db, graphroute, config, xmlnmap
+from ivre.analyzer import Context
```
p-l-: Can you `try` ... `except ImportError` this and set a global variable like `USE_ANALYZER`? This would make new dependencies "soft" (use `git grep USE_ bin/ ivre/` for examples).

ivre/analyzer.py
```diff
@@ -579,9 +625,12 @@ def search_anomalies(self, outliers, percent_mode=False):
...
         """
-        # 0.01 is faster and works well most of the time
-        # but 0.5 works with small host number.
-        nu_percent = outliers * 0.01 if percent_mode else 0.5
+        if percent_mode:
+            nu_percent = outliers * 0.01
+        else:
+            # 0.01 is faster but causes troubles with a small host number
+            # 0.5 works with a small host number but takes longer
+            nu_percent = 0.5 if len(self._list_id) < 500 else 0.01
```
p-l-: Would it make sense here to have, instead of just two different values, a value computed from `len(self._list_id)`?

Maybe something like `max(.01, -0.0005 * len(self._list_id) + 0.55)`, or `.5 * 10 ** (-float(len(self._list_id))/100)`?

ivre/tools/scancli.py
```diff
@@ -628,6 +655,24 @@ def displayfunction(x):
                     entry['_id'] = "None"
             print "%(_id)s: %(count)d" % entry
         sys.exit(0)
+    if USING_CONTEXT and args.tag_anomalies:
```
p-l-: One last thing: can you factorize `if USING_CONTEXT:` here?

p-l-:
Thanks!

It would be nice to add an item in the documentation files that describe the dependencies (at least `doc/README.md`, `doc/INSTALL.md` and `doc/FAST-INSTALL-AND-FIRST-RUN.md`), specifying that this is only needed for clustering / anomaly analysis.

Also, maybe the `docker/base*/Dockerfile`s could be updated to install the dependencies so that `ivre scancli` provides `--tag-*` options?

Last (but not least), a couple of tests could help! I'm OK to merge without the tests though, so that more people can test & use those features.

ivre/tools/scancli.py
```diff
@@ -476,6 +481,17 @@ def displayhost_csv(fields, separator, nastr, dic, out=sys.stdout):
                         for line in utils.doc2csv(dic, fields, nastr=nastr)))
     out.write('\n')
 
+def analyzer_init_context(hostfilter, skip=None, limit=None):
```
p-l-: For the sake of clarity, could you define this function only when `USING_CONTEXT`?

PR #162, opened 2015-10-02
Complete label implementation

p-l-:
After PR #153, complete support for labels. @xavier-martin, what do you think?


xavier-martin:
It looks great !

PR #191, opened 2015-12-03
KeyError catched

DariusMR:


ivre/tools/scancli.py
```diff
@@ -224,10 +224,13 @@ def display_honeyd_conf(host, honeyd_routes, honeyd_entries, out=sys.stdout):
...
-            hname, p['protocol'], p['port'],
-            nmap_port2honeyd_action(p))
-        )
+        try:
+            out.write('add %s %s port %d %s\n' % (
+                hname, p['protocol'], p['port'],
+                nmap_port2honeyd_action(p))
+            )
+        except KeyError:
+            pass
```
p-l-: Nice job! Maybe you could add a comment line before the `pass` statement explaining when this happens and why it is safe to ignore the exception and the port value leading to the exception?

PR #206, opened 2016-02-16
Added GeoIP infos update on Nmap DB when updating IP data

sp1d3rb0y:
Simple fix for updating GeoIP informations on already nmap-scanned hosts.

### DIFF
--- a/ivre/tools/ipdata.py
+++ b/ivre/tools/ipdata.py
@@ -66,7 +66,7 @@ def main():
                         help='Import FILE into locations database.')
     parser.add_argument('--import-all', action='store_true',
                         help='Import all files into databases.')
-    parser.add_argument('--dont-feed-ipdata-cols', action='store_true',
+    parser.add_argument('--no-update-passive-db', action='store_true',
                         help='Do not feed data to other purpose-specific'
                         ' DB (this only affects the passive DB for now)')
     parser.add_argument('--update-nmap-db', action='store_true',
@@ -95,10 +95,10 @@ def main():
     if args.download:
         ivre.geoiputils.download_all(verbose=not args.quiet)
     dbtofeed = {}
-    if not args.dont_feed_ipdata_cols:
-        dbtofeed["feedipdata"] = [ ivre.db.db.passive ]
-        if args.update_nmap_db:
-            dbtofeed["feedipdata"].append(ivre.db.db.nmap)
+    if not args.no_update_passive_db:
+        dbtofeed["feedipdata"] = [ivre.db.db.passive]
+    if args.update_nmap_db:
+        dbtofeed.setdefault("feedipdata", []).append(ivre.db.db.nmap)
     if args.city_csv is not None:
         TORUN.append((ivre.db.db.data.feed_geoip_city,
                       [args.city_csv],
--- a/ivre/tools/ipdata.py
+++ b/ivre/tools/ipdata.py
@@ -68,7 +68,7 @@ def main():
                         help='Import all files into databases.')
     parser.add_argument('--dont-feed-ipdata-cols', action='store_true',
                         help='Do not feed data to other purpose-specific'
-                        ' DB (this only affects the passive DB for now)')
+                        ' DB (this only affects the passive and nmap DB)')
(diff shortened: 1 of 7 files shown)


p-l-:
Hi,

Thanks for your contribution. The rationale behind the current choice of **not** updating Nmap results is that the information is kept as it was by the time of the scan, since a scan reflects the state of a host at a certain time (and, for that matter, seen from a certain source). The scan document includes `starttime` and `endtime` fields to reflect that.

What do you think?

sp1d3rb0y:
Hi,

Now I do understand the idea behind the "archives", and it makes sense to keep it that way.
However, in a situation where:
- My GeoIP database is ancient/obsolete ;
- My scans results have already been inserted ;
- I'd like to update my GeoIP information without re-scanning what I just scanned ;

This would be practical. Mainly when the scan date is more recent than the GeoIP database.

I do get that it's a really specific situation; maybe we could add it as an option to ipdata? (like '--update-scans-db')

p-l-:
Sorry for the delay. An `ivre ipdata` option `--update-nmap-db` would be perfect.

ivre/tools/ipdata.py
```diff
@@ -92,21 +94,23 @@ def my_parse_args():
                 ivre.db.db.data.ensure_indexes()
     if args.download:
         ivre.geoiputils.download_all(verbose=not args.quiet)
+    dbtofeed = {}
+    if not args.dont_feed_ipdata_cols:
+        dbtofeed["feedipdata"] = [ ivre.db.db.passive ]
+        if args.update_nmap_db:
```
p-l-: This one should probably be outside the `if not args.dont_feed_ipdata_cols:`: we may want to use both `--update-nmap-db` **and** `--dont-feed-ipdata-cols`, or do you think it does not make sense?

```python
dbtofeed = {}
if not args.dont_feed_ipdata_cols:
    dbtofeed["feedipdata"] = [ivre.db.db.passive]
if args.update_nmap_db:
    dbtofeed.setdefault("feedipdata", []).append(ivre.db.db.passive)
```

Also, please try to be [PEP8](https://www.python.org/dev/peps/pep-0008/)-compliant (no space inside brackets).

sp1d3rb0y: I assumed the purpose of the "dont-feed-ipdata-cols" option was "Update my GeoIP data, but do not update what has already been scanned, whether it's active or passive".

Maybe we could reverse the problem by removing the "dont-feed-ipdata-cols" option, replacing it by both an "update-passive-db" option and the "update-nmap-db" option.

This would make sense to explicitely specify whether to update one db or another, or both.

Would this solution do the trick?

I'll try to be PEP8-compliant from now on.

p-l-: The idea behind `--dont-feed-ipdata-cols` was that updating `passive` DB was the default behavior. Renaming it to `--no-update-passive-db`, to be consistent with your new option (and updating the help accordingly) would be OK to me. Thanks!

sp1d3rb0y: You do mean an "update-passive-db" option, or do you want to keep this default behavior?
In the latter case, I would probably rename the nmap option to "--no-update-nmap-db", since it would be confusing to have one non-negated option for one db, and a negated one for the other.

p-l-: No, I do mean to have `--no-update-passive-db` and `--update-nmap-db`, because I want the default to be: "update passive DB, don't update Nmap DB". That happens quite often, for example with `./configure` scripts.

ivre/tools/ipdata.py
```diff
@@ -66,9 +66,11 @@ def my_parse_args():
                         help='Import FILE into locations database.')
     parser.add_argument('--import-all', action='store_true',
                         help='Import all files into databases.')
-    parser.add_argument('--dont-feed-ipdata-cols', action='store_true',
+    parser.add_argument('--no-update-passive-db', action='store_true',
                         help='Do not feed data to other purpose-specific'
```
p-l-: You can rewrite the help as 'Do not update the passive database.'

PR #207, opened 2016-02-18
Replace screen by "proper" subprocess management in agent

fmonjalet:
The agent now spawns worker agent subprocesses, waits for them while at
least one is alive, and kills (SIGTERM) its whole process tree on exit,
SIGINT or SIGTERM.

This allows to use the agent more flexibly (for example in a systemd
service). As a bonus, it removes the `screen` dependency on the agent side, for what it is worth... 

The PR also adds some light logging to see what the agent workers are doing. @pl, could you also check that these messages are consistent?

As a preview/bonux, here is a sample systemd service file I use to run the agent on by boxes:

```cfg
[Unit]
Description=IVRE scan agent
After=network.target
Requires=network.target

[Service]
EnvironmentFile=-/etc/sysconfig/ivre-agent
WorkingDirectory=/var/lib/ivre-agent/
ExecStartPre=/bin/sh -c 'mkdir -p input output'
ExecStartPre=/bin/sh -c 'chgrp ivre input output'
ExecStartPre=/bin/sh -c 'chmod 2775 input'
ExecStartPre=/bin/sh -c 'chmod 2755 output'
# Replace by relevent path
ExecStart=/opt/ivre/agent/agent
User=root
Group=ivre

[Install]
WantedBy=multi-user.target
```

### DIFF
--- a/agent/agent
+++ b/agent/agent
@@ -1,7 +1,7 @@
 #! /bin/sh
 
 # This file is part of IVRE.
-# Copyright 2011 - 2015 Pierre LALET <pierre.lalet@cea.fr>
+# Copyright 2011 - 2016 Pierre LALET <pierre.lalet@cea.fr>
 #
 # IVRE is free software: you can redistribute it and/or modify it
 # under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


p-l-:
I see what you're doing here @fmonjalet. That's OK for this one, but be warned, I'm not going to let you remove all the dirty hacks from IVRE and replace them using clean and suited code.

PR #208, opened 2016-03-01
JSON export

p-l-:
This PR adds JSON export capability to the Web interface.


web/cgi-bin/scanjson.py
```diff
@@ -218,36 +224,41 @@ def main():
 
     version_mismatch = {}
     if callback is None:
-        sys.stdout.write("[\n")
+        tab, sep = "", "\n"
     else:
+        tab, sep = "\t", ",\n"
+        coma = ","
```
fmonjalet: Seems to be unused

PR #241, opened 2016-06-15
More precise time history for flows

fmonjalet:
This PR allows to record times when flows appear in IVRE Flows Neo4j db with a given granularity. The default config records every hour a flow appeared.

This allows to have a finer grained time handling than the current `first_seen`/`last_seen` and to make more precise requests (such as "Get me all the flows I have not seen in a month").


fmonjalet:
Seems ready to be merged for me. I switched from a time tree to simple timestamp (Time) nodes, simpler, faster, more flexible and easier to handle.

PR #250, opened 2016-06-23
WebUI/Flow: timeline highlight works with nodes

p-l-:


web/static/ivre/flow/controllers.js
```diff
@@ -583,16 +578,29 @@ ivreWebUi
 
         };
 
-        $scope.timeline_highlight_flow = function (flow_id) {
+        $scope.timeline_highlight_flow = function (elt) {
             var time_prec = config.flow_time_precision * 1000;
-            if (flow_id === undefined && $scope.cur_elt !== undefined) {
-                flow_id = $scope.cur_elt.id;
+	    if (elt === undefined) {
```
fmonjalet: Smells like a tab. :)

PR #267, opened 2016-07-13
Partial parsing of mongodb-{info,databases} NSE scripts

commial:
This PR:
* Add error magics for [`mongodb-info`](https://nmap.org/nsedoc/scripts/mongodb-info.html), [`mongodb-databases`](https://nmap.org/nsedoc/scripts/mongodb-databases.html) NSE scripts.
* Parse `mongodb-databases` output, including databases and shards.
* Fix the order of execution of output error detection and output parsing.

### DIFF
--- a/ivre/db/mongo.py
+++ b/ivre/db/mongo.py
@@ -411,6 +411,8 @@ class MongoDBNmap(MongoDB, DBNmap):
                   "ports.scripts.smb-enum-shares.shares",
                   "ports.scripts.ls.volumes",
                   "ports.scripts.ls.volumes.files",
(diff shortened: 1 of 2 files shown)


commial:
Actually, it seems that on the SVN version of Nmap, the different numbers are reported as float.
(The PR has been tested with Nmap 7.12, the latest stable).

Should I replace `int` parsing with `float` (as `float` can handle both `0` and `0.0`) or do we wait the official release (it may change since then)?

p-l-:
I think we should have `scan2db` work even against the SVN version. Using `float` seems OK for me.

commial:
Updated and tested.

PR #272, opened 2016-07-25
Multiple improvements for active recon

p-l-:
This PR groups several improvements made for active recon after running wide Internet scans with more Nmap options than before.

See each commit for the detail.

@commial @fmonjalet @carolineLe can you give your opinion?


ivre/xmlnmap.py
```diff
@@ -1282,7 +1282,7 @@ def __init__(self, fname, categories=None, source=None,
         self._collection = self._db.nmap.db[self._db.nmap.colname_hosts]
         self._scancollection = self._db.nmap.db[self._db.nmap.colname_scans]
         if gettoarchive is None:
-            self._gettoarchive = lambda c, a, s: []
+            self._gettoarchive = lambda a, s: []
```
commial: This is not explicitly related to this commit, but it seems that `gettoarchive` is used as a keyword input to `NmapHandler`, which actually ignores it.

ivre/db/mongo.py
```diff
@@ -182,6 +182,12 @@ def migrate_schema(self, colname, version):
             updated = False
             new_version, migration_function = self.schema_migrations[
                 colname][version]
+            if config.DEBUG:
```
commial: The `logging` module seems more adapted for this (but it will certainly need a bigger patch).

ivre/xmlnmap.py
p-l-: That's only so that `NmapHandler` and `Nmap2Mongo` constructors have the same prototype.

ivre/db/mongo.py
p-l-: Agreed, `logging` should be used everywhere, but that's another job! Maybe you could open an issue to track this?

commial: Done in #273

web/static/ivre/controllers.js
```diff
@@ -380,13 +380,16 @@ ivreWebUi
...
+		 return "label-warning";
 	    }
 	};
 	$scope.short_port_status = function(status) {
-	    if(status === "filtered")
-		return "fltred";
-	    return status;
+	    return {
+		"filtered": "fltred",
+		"open|filtered": "opn | flt"
```
fmonjalet: This remark is purely esthetic, but:
- filtered is mapped to fltred (why not plain "filtered" or just "flt"?)
- but `open|filtered` is mapped to `opn | flt` and not `opn | fltred`, or better `opn|fltred`.
Maybe you could switch to `flt` and `opn|flt`? I can't see the impactof this, though.

p-l-:
Thanks everyone!

PR #275, opened 2016-08-03
DB/Mongo: add .distinct() method using the aggregation framework

### LINKED ISSUE #274: scancli: use of --limit, --skip and --short are mutually exclusive
Hi,

There is an use case with `scancli` when you want to extract an IP subset of your dataset, so you may want to use `--limit` combined with `--short`.
In this case result is unattended: you receive the whole IP list of your dataset.

By looking at the code, you use the `distinct()` method that is ignoring the `limit()` set previously on the db cursor.

In order to get the expected behavior I changed the code like this:

```
diff --git a/ivre/tools/scancli.py b/ivre/tools/scancli.py
index 6b82493..90ae0a7 100644
--- a/ivre/tools/scancli.py
+++ b/ivre/tools/scancli.py
@@ -531,11 +531,11 @@ def main():
                                  default=db.db.nmap.serialize)
     elif args.short:
         def displayfunction(x):
-            for h in x.distinct('addr'):
+            for h in x:
                 try:
-                    out.write(utils.int2ip(h) + '\n')
+                    out.write(utils.int2ip(h['addr']) + '\n')
                 except:
-                    out.write(str(h) + '\n')
+                    out.write(str(h['addr']) + '\n')
     elif args.honeyd:
         def displayfunction(x):
             display_honeyd_preamble(out)
```

That modification is still not satisfying me because of the  performance impact.
I have ~40k entries, and got 0.4sec vs. 1.4sec (original code, patched code) by doing a simple `ivre scancli --short`...

Cheers dans la casa

p-l-:
This saves memory and is much more efficient.

Also, `ivre scancli` uses `.distinct()`, which fixes #274.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,4 +49,4 @@ before_script:
   - wget -O - "https://wiki.wireshark.org/SampleCaptures?action=AttachFile&do=get&target=imap.cap.gz" | zcat > tests/samples/imap.pcap
   - for file in telnet-raw.pcap nb6-startup.pcap nb6-http.pcap nb6-telephone.pcap nb6-hotspot.pcap ; do wget "https://wiki.wireshark.org/SampleCaptures?action=AttachFile&do=get&target=$file" -O "tests/samples/$file" ; done
 
-script: cd tests/ && python tests.py
+script: cd tests/ && travis_wait 30 python tests.py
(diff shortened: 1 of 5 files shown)


p-l-:
@fmonjalet or @commial, can you review this one?

PR #293, opened 2016-09-01
Make --nmap-prescan-ports functionnal

CyrilleFranchet:
If we use --nmap-prescan-ports without the --nmap-prescan-opts option, target.py will try to split the opts string with shlex.split() which is an incorrect action.


p-l-:
Thanks, nice catch!

PR #423, opened 2017-10-11
Update __init__.py

r00t0vi4:
Added no-timeago argument


p-l-:
Can you redo this using a PR title (and commit message) that explains what you did rather than "Update __init__.py"?

PR #430, opened 2017-11-17
Additional changes if mongo is fresh installed

SeppBender:
Just some information for simpler installation.


p-l-:
This is not needed for IVRE to work. Moreover, the steps should be taken from MongoDB's documentation, not from IVRE (IVRE can use PostgreSQL instead of MongoDB).

PR #431, opened 2017-11-17
Doc to install py2neo package

SeppBender:
Documented to use py2neo package for using neo4j

### DIFF
--- a/doc/FLOW.md
+++ b/doc/FLOW.md
@@ -11,8 +11,11 @@ It can be seen as:
 
 # Installation #
 
-Start by following the standard ivre installation.
-
+Start by following the standard ivre installation. Don't forget to install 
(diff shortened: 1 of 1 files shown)


doc/FLOW.md
```diff
@@ -11,8 +11,11 @@ It can be seen as:
 
 # Installation #
 
-Start by following the standard ivre installation.
-
+Start by following the standard ivre installation. Don't forget to install 
+py2neo package
```
p-l-: I'd be OK to make it clear here that py2neo (and neo4j, actually) is required. However, I would not recommend one particular method to install it over others. Could you remove the script (and mention the installation of a Neo4j server)?

p-l-:
No activity.

PR #438, opened 2017-11-24
ipinfo: fix `--short` when an IP is None

fmonjalet:
Also do not display `None` value when doing `ivre ipinfo --distinct addr`

### DIFF
--- a/ivre/tools/ipinfo.py
+++ b/ivre/tools/ipinfo.py
@@ -127,7 +127,8 @@ def disp_recs_std(flt):
 
 def disp_recs_short(flt):
     for addr in db.passive.distinct('addr', flt=flt):
-        print(ivre.utils.int2ip(addr))
+        if addr is not None:
+            print(ivre.utils.int2ip(addr))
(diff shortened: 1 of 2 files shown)


ivre/tools/ipinfo.py
```diff
@@ -127,12 +127,14 @@ def disp_recs_std(flt):
...
     for addr in db.passive.distinct('addr', flt=flt):
-        print(ivre.utils.int2ip(addr))
+        if addr is not None:
+            print(ivre.utils.int2ip(addr))
 
 
 def disp_recs_distinct(field, flt):
     for value in db.passive.distinct(field, flt=flt):
-        print(value)
+        if value is not None:
```
p-l-: I'm not sure about this one. Maybe it's good to know that `None` is a possible value?

fmonjalet: Okay, let's remove it then!

fmonjalet:
Fixed by #441

PR #440, opened 2017-12-04
Add set[..] parsing in BroFile

carolineLe:


p-l-:
Thanks! We may also want to handle `table[subtype]` types. What do you think?

carolineLe:
Yes, I added the three container types (set, vector and table).

PR #441, opened 2017-12-12
Fix address type mismatching for ipinfo --short

Nutigoodpipes:


p-l-:
Looks better now. Care to add a unit test? Maybe you could borrow it from #438?

PR #467, opened 2018-01-21
feat(docker/client): install  python-pil needed by scan2db to trim_image 

fccagou:
avoid ivre/utils.py:702 warning


p-l-:
Thanks @fccagou!

PR #477, opened 2018-02-11
feat (parser): add parser for iptables sylog

fccagou:
New parser for iptables log.

At the begining of iptables.py, can you have a look to the questions in the 'TODO:' section.
I will code it depending of your point of view for the next version of the code.

++


ivre/parser/iptables.py
```diff
@@ -0,0 +1,81 @@
...
+        super(Iptables, self).__init__(fname)
+
+
+    @classmethod
+    def parse_line(self, line):
+        line = line.decode().rstrip('\r\n')
+        if not line:
+            return next(self)
+
+        fields={}
```
p-l-: I prefer using comprehension expressions when possible. Don't you think this would work?
```python
fields = dict(val.split('=', 1) if '=' in val else (val, "") for val in line[line.find('IN='):].split())
```

Also, maybe it would be better to check if 'IN=' appears in the line (and return `next(self)` if it does not). What do you think?

And, given the following lines, maybe we could use the opportunity to `.lower()` the keys:

```python
fields = dict((key.lower(), value) for key, value in (val.split('=', 1) if '=' in val else (val, "") for val in line[line.find('IN='):].split()))
```

ivre/parser/iptables.py
```diff
@@ -0,0 +1,81 @@
...
+# along with IVRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# Maintainers:
+#   - 2018 Francois CHENAIS <francois.chenais@cea.fr>
+#
+
+"""Support for Iptables log from syslog files
+
+
+Feb  4 05:30:11 pi01 kernel: [3240403.495065] IPTABLES/UDP/IN=enxb827eb8f8a4f OUT= MAC=ff:ff:ff:ff:ff:ff:18:62:2c:7e:45:d0:08:00:45:00:00:ec:00:00:40:00:40:11:df:e3 SRC=192.168.0.254 DST=192.168.0.255 LEN=236 TOS=0x00 PREC=0x00 TTL=64 ID=0 DF PROTO=UDP SPT=138 DPT=138 LEN=216
```
p-l-: I think the prefix (`IPTABLES/UDP/IN=`) should not be hard-coded but somehow configurable.

ivre/parser/iptables.py
```diff
@@ -0,0 +1,81 @@
...
+
+        # Rename fields according to flow2db specifications.
+        fields['src'] = fields.pop('SRC')
+        fields['dst'] = fields.pop('DST')
+        fields['proto'] = fields.pop('PROTO').lower()
+        if fields['proto'] in ('udp', 'tcp'):
+            fields['sport'] = int(fields.pop('SPT'))
+            fields['dport'] = int(fields.pop('DPT'))
+
+        # This data are mandatory but undefined in iptables logs, so make my/a choice.
```
p-l-: **is** mandatory ;-)

ivre/parser/iptables.py
fccagou: The prefix depends of the iptables conf  --log-prefix. In this case the log prefix is 'IPTABLES/UDP/' as an example. Because the log isn't separate from the first field IN=, I use the find at the begin of the parse_line function.
It should be safer to find the last occurancy if the prefix contains IN=...:P

ivre/parser/iptables.py
fccagou: ho! ho! 
What's the best way to process proto without sport or dport ?

p-l-: I don't get your question. You mean when proto is ICMP for example? I **think** we use type and code for sport & dport, but I'm not really sure. Maybe @fmonjalet knows?

ivre/parser/iptables.py
p-l-: Yep I understand that. Maybe we could by default try to match any iptables log (without caring about the prefix) and propose an option to look for a specific prefix?

ivre/parser/iptables.py
fccagou: For the return condition, I use the example of argus and suppose that the input file contains only netfilter logs according to the FAQ example.  Your suggestion would allow parsing files without previous filter processing. It's fine !

For the choice of comprehension expression, U are the master, I do like the master want !:D 

I do the change.

ivre/parser/iptables.py
fccagou: Yes.

fccagou: Hum, seems working yet without dport and sport ...

fccagou: Proto is used in IGMP example.

fccagou: OK ... I just realize you were talking about the conjugation 🤣 !!

fccagou:
I've push the fix.

I realize we should make a syslog parser collecting all type of data.
dhcp for port 67, mail for smtp/pop/imap, sshd, dns ...

Could it be pertinent to say that a dns query is mostly followed by a connection. The port and protocol are unknown but it could be marked by a specific tag for queries.

ivre/parser/iptables.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # Converts the syslog iptables log into hash
+        fields = dict((key.lower(), value) 
+                for key, value in (val.split('=', 1) if '=' in val else (val, '') 
+                    for val in line[field_idx:].rstrip('\r\n').split()))
+
+        # Because day of month can be on 1 or 2 char(s)
+        try:
+            fields['start_time'] = datetime.datetime.strptime(line[:15], '%b  %d %H:%M:%S')
+        except:
```
p-l-: It's better to narrow the exception you want to catch with this `except`. Anyway, I **think** using `"%w"` should work in both cases:
```python
    fields['start_time'] = datetime.datetime.strptime(line[:15], '%b %w %H:%M:%S')
```

fccagou: I'll test it.
I also wanted to add some non reg tests.
I think about creating tests/samples  directory with iptables.log file containing the sample file. 
This dir could then contains all types of samples.
What do you think ?

fccagou: datetime.datetime.strptime(line[:15], '%b %w %H:%M:%S')  doesn't woks when the day number contains 2 digits.

datetime.datetime.strptime(line[:15], '%b %d %H:%M:%S')  works in both cases. Huummmm,
I thought I had an exception during my tests ... I'll had the case in the sample.

p-l-: That would be great! For now, because @fmonjalet is lazy, we have no flow regression tests. You could at least test that your parser works in `test_utils` for example.

fccagou: Tests in progress....https://travis-ci.org/fccagou/ivre/

tests/tests.py
```diff
@@ -1612,6 +1612,20 @@ def is_prime(n):
                         i + 1,
                     )
 
+    def test_parser(self):
```
p-l-: Could you rather move this at the end of `test_utils()`, after the Bro parser tests?

fccagou: Done and tests are ok.

Issue #319, opened 2016-10-09
Search Bro info from web console


fjavierc:
I would like to see the information collected by Bro from the Ivre web console in the same way that I do it through "ivre ipinfo".

 Some queries  would be, for example:

- "**1.2.3.4**" for all information collected by Bro for IP 1.2.3.4 (as run  ' _ivre ipinfo 1.2.3.4_ ')
- "**pop**" to obtain information from the database in which " recontype = POP_CLIENT " or " recontype = POP_SERVER"  (as run ' _ivre ipinfo --pop_ ')
- "**pop 1.2.3.4**" to obtain information from the database in which "recontype = POP_CLIENT" or "recontype = POP_SERVER"  but also associated with ip address 1.2.3.4  ( as run ' _ivre ipinfo --pop 1.2.3.4_ ')

I  could add a new form and include a filter but ... I do not know how to continue, I do not know how to return the results.
![image1](https://cloud.githubusercontent.com/assets/15417469/19222524/de25be76-8e5a-11e6-8db9-adb50da0f582.png)  
![image2](https://cloud.githubusercontent.com/assets/15417469/19222526/e54613cc-8e5a-11e6-9b09-5e5ec9aa8f01.png)

Would it be very difficult to do? Is there a base template that I can use ?

Thanks

p-l-:
Hi!

Sorry but there is no Web interface for the "passive" database for now. We have plans to merge the information from "passive", "nmap" and "flow", but that's not for a near future.

I'll let this issue open so that other people can find it and maybe someone will want to implement it!

p-l-:
You can now build a "view" from the passive database (run `ivre db2view passive`) and then use the Web UI to browse it. Please note that not all passive data types are supported for now.

PR #492, opened 2018-03-07
Support "view" feature

Nutigoodpipes:
This PR adds a new feature, called "view", with "nmap", "passive" and "data". This feature will be used by the web interface, and will receive data from both "nmap" and "passive". Its structure is very close to "nmap".


p-l-:
This would fix #319.

doc/README.md
```diff
@@ -243,6 +241,27 @@ To use the Python module, run for example:
 
 For more, run `help(db.nmap)` from the Python shell.
 
+# Create a view #
+
+## Import results ##
+
+You can import results from both passive and nmap databases. To create
+a view using every results just use :
```
p-l-: using **all the results in both database,** just use

doc/README.md
```diff
@@ -243,6 +241,27 @@ To use the Python module, run for example:
...
+
+To give filters you can use the `--filters` options. For more infos
+just run `ivre help db2view` from the shell.
+
+Warning: if you don't specify either `--use-nmap` or `--use-passive`,
+both will be used, and no filters means all results.
+
+## Using the view ##
+
+Once inserted, you can browse results as you would do with `scancli`
```
p-l-: "you can browse results with \`viewcli\` as you would do with \`scancli\`"?

PR #550, opened 2018-07-13
runscansagentdb assign targets with an agent_id.

TiTi87:
`Runscansagentdb --assign` can now directly assign targets to an agent. This permit to avoid specifying a scan without agent and then specify the agent for the scan on another command.

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -1811,7 +1811,7 @@ which `predicate()` is True, given `webflt`.
             proc.kill()
             proc.wait()
             signal.alarm(old_alarm)
-        
+
         signal.signal(signal.SIGALRM, old_handler)
 
         if DATABASE == "mongo":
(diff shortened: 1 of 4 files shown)


tests/tests.py
```diff
@@ -2119,41 +2119,41 @@ def is_prime(n):
...
-        with AgentScanner(self) as agent:
-            agent.scan(["--test", "2"])
-
-        # Count the results
-        count = sum(len(walk_elt[2]) for walk_elt in os.walk('output/'))
-        self.assertEqual(count, 2)
-
-        # Clean
-        shutil.rmtree('output')
+        # # Check simple runscans
```
p-l-: Why did you comment those lines?

TiTi87: Sorry my mistake, was working on the tests and I did'nt realize my pull request will update here to...
I though i was working on my fork.... I think I'll close this pull request and put a proper one once done...

PR #556, opened 2018-07-20
Add docker compose support

Edznux:
Hello,

I added support for docker-compose as it's a more common way to deploy simple application with docker.
It adds the auto restart of the whole infrastructure and could easily be modified to set some restriction (cpu/disk/mem...)
I commented the "v3" thing just like in the Vagrantfile :)

Thanks


Edznux:
Done :)

PR #558, opened 2018-07-24
Update Vagrantfile

Maximevilla:
remove bug on windows for portForward in VagrantFile


p-l-:
Unless I'm wrong, this is not related to a bug in Windows. This should just choose another port in case port 80 is already in use on your system. Am I correct?

p-l-:
I'm closing this for no activity. Please add a comment if it needs to be considered.

PR #600, opened 2018-11-08
Fix minor things v1

serializingme:
This PR fixes two minor things:
- The `view-ports-only.html` (used by the `displayPort` directive) and `view-services-only.html` (used by the `displayService` directive) files are missing from the `setup.py` file (since its creation).
- When an host doesn't have a name, some "OCD nagging" empty parentheses `()` are left behind in the rendered `subview-host-summary.html` template.


serializingme:
Another minor fix on the way, will close this PR and open another one.

PR #601, opened 2018-11-08
WebUI: multiple fixes

serializingme:
This PR fixes three minor things:
- The `view-ports-only.html` (used by the `displayPort` directive) and `view-services-only.html` (used by the `displayService` directive) files are missing from the `setup.py` file (since its creation).
- When an host doesn't have a name, some "OCD nagging" empty parentheses `()` are left behind in the rendered `subview-host-summary.html` template.
- Using `ng-src` instead of `src` allowing Angular to compile the template 
before the browser tries to download the screenshot (shown in the browser console as an error `GET data:image/jpeg;base64,{{::port.screendata}} net::ERR_INVALID_URL`)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -188,6 +188,8 @@ specialized scripts.
           'web/static/templates/view-hosts.html',
           'web/static/templates/view-screenshots-only.html',
           'web/static/templates/view-scripts-only.html',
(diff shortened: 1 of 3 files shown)


p-l-:
Thanks a lot for this contribution!

Issue #618, opened 2018-12-02
Passive: hide `fullvalue` & `fullinfos` from the user


p-l-:

p-l-:
Fixed by #622 & #627.

PR #622, opened 2018-12-04
Passive: hide full{value,infos} from user, move it to MongoDB specific code

Nutigoodpipes:
Move 'fullvalue' and 'fullinfos' functions in mongodb specific code, since they are used only for size restrictions on mongodb indexes.

### DIFF
--- a/ivre/db/mongo.py
+++ b/ivre/db/mongo.py
@@ -30,6 +30,7 @@ except ImportError:
     OrderedDict = dict
 from copy import deepcopy
 import datetime
+import hashlib
 import json
 import os
 import re
@@ -3351,6 +3352,11 @@ is not expected)."""
                                                 rec.pop('addr_1')])
             except (KeyError, socket.error):
                 pass
+            for key in ['value', 'targetval']:
+                if 'full' + key in rec:
+                    rec[key] = rec.pop('full' + key)
+            if 'fullinfos' in rec:
+                rec.setdefault('infos', {}).update(rec.pop('fullinfos'))
             yield rec
 
     def get_one(self, spec, **kargs):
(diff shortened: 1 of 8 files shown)


p-l-:
@Nutigoodpipes do you have two commits with the same title on purpose?

ivre/db/mongo.py
```diff
@@ -3351,6 +3352,11 @@ def get(self, spec, hint=None, **kargs):
                                                 rec.pop('addr_1')])
             except (KeyError, socket.error):
                 pass
+            for key in ['value', 'targetval']:
+                if 'full' + key in rec:
+                    rec[key] = rec.pop('full' + key)
+            if 'fullinfos' in rec:
+                rec.setdefault('infos', {}).update(rec['fullinfos'])
```
p-l-: Maybe we even want to `.pop('fullinfos')`?

ivre/db/mongo.py
```diff
@@ -3368,6 +3374,11 @@ def get_one(self, spec, **kargs):
             pass
         else:
             del rec['addr_0'], rec['addr_1']
+        for key in ['value', 'targetval']:
+            if 'full' + key in rec:
+                rec[key] = rec.pop('full' + key)
+        if 'fullinfos' in rec:
+            rec.setdefault('infos', {}).update(rec['fullinfos'])
```
p-l-: Same here.

ivre/db/mongo.py
```diff
@@ -3376,6 +3387,32 @@ def update(self, spec, **kargs):
...
+                spec[key] = hashlib.sha1(spec[key].encode()).hexdigest()
+
+    @staticmethod
+    def _fix_infos_size(spec):
+        # We enforce a utils.MAXVALLEN // 10 size limits for subkey values in
+        # infos; this is because MongoDB cannot index values longer than 1024
+        # bytes.
+        for field in list(spec.get("infos", {})):
+            value = spec["infos"]
+            if len(value) > utils.MAXVALLEN:
```
p-l-: That does not seem consistent with the above comment (`utils.MAXVALLEN // 10`?).

p-l-: BTW, is there a reason to have two distinct methods here?

p-l-:
BTW this would fix #618.

ivre/db/mongo.py
Nutigoodpipes: Same reason I had two commits, uncleaned dev process...

Nutigoodpipes: Wrong fix now fixed.

PR #631, opened 2018-12-21
Nmap: Ignore sslv2-drown NSE script

serializingme:
Since the sslv2-drown NSE script mixes elem/table tags with and without key attributes which is currently unsupported, the script needs to be ignored. This was causing the import of some hosts to fail.

### DIFF
--- a/ivre/xmlnmap.py
+++ b/ivre/xmlnmap.py
@@ -47,7 +47,7 @@ SCHEMA_VERSION = 11
 
 # Scripts that mix elem/table tags with and without key attributes,
 # which is not supported for now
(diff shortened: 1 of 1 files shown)


p-l-:
Hi,

Thanks for this patch! Could you by any chance share an example of file causing a failure?

serializingme:
The problem seems to be the `ciphers` table that just has a list of `elem` without identifiers. You can find an example [here](https://gist.github.com/serializingme/03b3f1308cc94e48dcb7823808d8011b).

p-l-:
Thanks a lot!

PR #634, opened 2018-12-26
WebUI: Added display vulnerabilities only filter.

serializingme:
Updated the previous pull request (#633) based on the comments received. Even though it didn't seem necessary, to be on the safe side I added the `vulns !== undefined` check.

### DIFF
--- a/web/static/ivre/controllers.js
+++ b/web/static/ivre/controllers.js
@@ -328,8 +328,10 @@ ivreWebUi
 	$scope.results = [];
 	$scope.display_mode = "host";
 	$scope.display_mode_args = [];
-	$scope.script_display_mode_needed_scripts_group = function(scripts) {
+	$scope.script_display_mode_needed_scripts_group = function(scripts, vulns) {
 	    if(scripts === undefined || scripts.length === 0)
+		return false;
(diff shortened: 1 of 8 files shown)


web/static/ivre/controllers.js
```diff
@@ -328,9 +328,11 @@ ivreWebUi
 	$scope.results = [];
 	$scope.display_mode = "host";
 	$scope.display_mode_args = [];
-	$scope.script_display_mode_needed_scripts_group = function(scripts) {
+	$scope.script_display_mode_needed_scripts_group = function(scripts, vulns) {
 	    if(scripts === undefined || scripts.length === 0)
 		return false;
+	    if(vulns !== undefined && vulns && !scripts.some(function (x) { return x.hasOwnProperty('vulns'); }))
```
serializingme: @p-l- Looking a second time on this one, `vulns === true` should suffice  and it would be shorter than `vulns !== undefined && vulns`. Which one do you prefer? I will redo the PR, sorry about that.

p-l-: Both are fine to me. Your code, your call! ;-)

Issue #642, opened 2019-01-05
Top values broken for JSON fields with SQLite backend


p-l-:
Please try to follow this template. Also, please format your issue
properly (particularly to report code or command line output). You may
want to read
[Github's basic writing and formatting syntax](https://help.github.com/articles/basic-writing-and-formatting-syntax/).

### IVRE Version

Current development version (`0.9.11.dev_master_211`).

### Summary

`.topvalues()` is broken for JSON subfields with SQLite backend.

### Expected behavior

```
$ ivre ipinfo --top infos.service_product --service ftp
Pure-FTPd: 1
```

### Actual behavior

```
$ ivre ipinfo --top infos.service_product --service ftp
Traceback (most recent call last):
  File "/usr/bin/ivre", line 84, in <module>
    main()
  File "/usr/bin/ivre", line 56, in main
    tools.get_command(possible_commands[0])()
  File "/usr/lib/python3.7/site-packages/ivre/tools/ipinfo.py", line 338, in main
    args.skip or 0)
  File "/usr/lib/python3.7/site-packages/ivre/tools/ipinfo.py", line 163, in <lambda>
    limit)
  File "/usr/lib/python3.7/site-packages/ivre/utils.py", line 1558, in display_top
    for entry in db.topvalues(field, flt=flt, topnbr=lmt or 10, least=least):
  File "/usr/lib/python3.7/site-packages/ivre/db/sql/__init__.py", line 1901, in <genexpr>
    for result in self.db.execute(req.order_by(order).limit(topnbr))
  File "/usr/lib/python3.7/site-packages/ivre/db/sql/tables.py", line 138, in process
    value = json.loads(value)
  File "/usr/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

### How to reproduce

Use an SQLite database, insert the results used for the Travis CI tests and run `ivre ipinfo --top infos.service_product --service ftp`.

### More

When this bug is fixed, a new test can be added (the output of `ivre ipinfo --top infos.service_product --service ftp` for example).

The following code block in the test should also be removed:
[...]

PR #646, opened 2019-01-12
Fix topvalues for passive with sqlite backend.

Nutigoodpipes:
Fields for passive table are now raw json instead of astext.
This is transparent for postgresql, but for some reason data
are processed again with sqlite backend. (This may be due to
JSON type not being native for this backend as sqlite-3.26).


Nutigoodpipes:
Would fix #642.

ivre/db/sql/tables.py
```diff
@@ -79,10 +79,18 @@ def access(d, k):
...
-        return str(json.loads(d).get(k))
-
-    dbapi_connection.create_function('ACCESS_TXT', 2, access_astext)
+    # FIXME: astext as a manual string cast for JSON.
+    # astext is dirty here because strings value in JSON are not
+    # properly converted to themselves. This may be due to improper
+    # type handling with the custom SQLJSONB.
+    # Proper way to do would be : `cast(x, String)`
+    def astext(x):
+        r = str(x)
```
p-l-: Wouldn't this work:
```Python
        return str(json.loads(r))
```

Nutigoodpipes: Yes, it would. At first, this line ```r = str(x)``` is useless because argument is a json field so a string already.

The why of ```r.startswith('"')``` is for consistency with simple and double quote.
```>>> str(json.loads('{"1": 1}'))
"{'1': 1}"
>>> str('{"1": 1}')
'{"1": 1}'
```
The last one is still a json, even if astext does not guarantee that output format.

PR #664, opened 2019-03-01
DB/View: support regexp lookups in .searchja3*()

p-l-:
This PR also fixes `DBView.searchja3server()` API to match `DBPassive.searchja3server()`.

### DIFF
--- a/ivre/db/__init__.py
+++ b/ivre/db/__init__.py
@@ -1214,15 +1214,19 @@ class DBView(DBActive):
     def parse_args(self, args, flt=None):
         flt = super(DBView, self).parse_args(args, flt=flt)
         if args.ssl_ja3_client is not None:
+            cli = args.ssl_ja3_client
             flt = self.flt_and(flt, self.searchja3client(
-                value_or_hash=args.ssl_ja3_client
+                value_or_hash=(
+                    False if cli is False else utils.str2regexp(cli)
+                )
             ))
         if args.ssl_ja3_server is not None:
             if args.ssl_ja3_server is False:
                 # There are no additional arguments
                 flt = self.flt_and(flt, self.searchja3server())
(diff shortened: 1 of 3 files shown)


ivre/web/utils.py
```diff
@@ -472,26 +472,32 @@ def add_unused(neg, param, value):
...
                 neg=neg
             ))
         elif param == "ssl-ja3-server":
             if value is None:
                 # There are no additional arguments
                 flt = db.view.flt_and(flt, db.view.searchja3server(neg=neg))
             else:
-                splitted = value.split(':', 1)
-                if len(splitted) == 1:
+                split = [v if v else None for v in value.split(':', 1)]
```
vruello: Why don't you apply `utils.str2regexp` here instead of line 490/498/500 ? Like : 
``` 
split = [utils.str2regexp(v) if v else None for v in value.split(':', 1)]
``` 
If so, you could remove the if conditions line 489, 497 and 499 and give directly variables `srv` and `cli` .

ivre/db/__init__.py
```diff
@@ -1215,24 +1215,25 @@ def parse_args(self, args, flt=None):
...
+                               utils.str2regexp(args.ssl_ja3_client))
             ))
         if args.ssl_ja3_server is not None:
             if args.ssl_ja3_server is False:
                 # There are no additional arguments
                 flt = self.flt_and(flt, self.searchja3server())
             else:
-                splitted = args.ssl_ja3_server.split(':', 1)
-                if len(splitted) == 1:
+                split = args.ssl_ja3_server.split(':', 1)
```
vruello: Maybe you could apply `utils.str2regexp` on every elements of the array instead of manually line 1230, 1235 and 1236.

PR #666, opened 2019-03-04
Import User-Agents from passive into view with corresponding filters

vruello:
The option `http-user-agent` (cli and webui) accepts : 
- nothing : every hosts known to use a User-Agent will be retrieved.
- or a string : that will retrieve every hosts which use this exact User-Agent
- or a regular expression : that will retrieve every hosts which use a User-Agent that matches the regex

### DIFF
--- a/ivre/utils.py
+++ b/ivre/utils.py
@@ -312,7 +312,7 @@ def str2regexp(string):
 
     """
     if string.startswith('/'):
-        string = string.split('/', 2)[1:]
+        string = string[1:].rsplit('/', 1)
         if len(string) == 1:
             string.append('')
         string = re.compile(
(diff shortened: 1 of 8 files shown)


p-l-:
The SQLite tests should be OK now, so you can rebase against current master and push again.

ivre/utils.py
```diff
@@ -312,7 +312,7 @@ def str2regexp(string):
 
     """
     if string.startswith('/'):
-        string = string.split('/', 2)[1:]
+        string = string[1:].rsplit('/', 1)
```
p-l-: If we want to accept `/` in regexp, they should be escaped. We probably want to replace `\/` by `/`.

PR #669, opened 2019-03-06
Enable slash-escape in regexp

vruello:


ivre/utils.py
```diff
@@ -305,20 +305,50 @@ def get_domains(name):
     return ('.'.join(name[i:]) for i in range(len(name)))
 
 
+def espace_slash(string):
+    """This function transforms '\/' in '/'' if '\' was not escaped
```
p-l-:
```suggestion
    """This function transforms '\\/' in '/' but leaves '\\\\/' unchanged. This is
useful to parse regexps from JavaScript style (/regexp/).
```

ivre/utils.py
```diff
@@ -305,20 +305,50 @@ def get_domains(name):
...
+    """  # noqa: W605
```
p-l-:
```suggestion
    """
```

ivre/utils.py
```diff
@@ -305,20 +305,50 @@ def get_domains(name):
     return ('.'.join(name[i:]) for i in range(len(name)))
 
 
+def espace_slash(string):
```
p-l-:
```suggestion
def _escape_slash(string):
```

ivre/utils.py
```diff
@@ -305,20 +305,50 @@ def get_domains(name):
...
 def str2regexp(string):
     """This function takes a string and returns either this string or
     a python regexp object, when the string is using the syntax
     /regexp[/flags].
-
     """
     if string.startswith('/'):
         string = string[1:].rsplit('/', 1)
+        # Enable slash-escape even if it is not necessary
+        string[0] = espace_slash(string[0])
```
p-l-:
```suggestion
        string[0] = _escape_slash(string[0])
```

ivre/utils.py
```diff
@@ -305,20 +305,50 @@ def get_domains(name):
...
+        elif escaping and char != '/':
+            new_string += '\\' + char
+            escaping = False
+        else:
+            new_string += char
+            escaping = False
+    return new_string
+
+
+def escape_first_slash(string):
```
p-l-: I don't get the purpose of this function.

ivre/web/utils.py
```diff
@@ -138,7 +137,7 @@ def query_from_params(params):
             else [neg, pval, None]
             for neg, pval in (
                 (True, x[1:]) if x[:1] in '!-' else (False, x)
-                for x in shlex.split(query)
+                for x in query.split()
```
p-l-: Is this related to this PR?

ivre/utils.py
vruello: If one wants to search for the string `/test`, one would use `\/test` so that the first `/` does not make the string interpreted as a regular expression. Then, it is needed to remove this extra `\`.

ivre/web/utils.py
vruello: Yes! `shlex` causes strange behaviors in web filters. For example, if one wants to match the regex `\d*`, one would use `<something>:/\d*/`. `shlex.split()` interprets the given string as a "shell-like" command and transforms `\d` in `d`. Thus, the pattern checked is `/d*/` which has a significantly different meaning. Currently, the workaround is to escape `\` like `/\\d*/`, which does not seem user friendly to me.

p-l-: I get it. The problem here is that `shlex.split()` is quite different than `str.split()`:
```
In [1]: "m1 'm2 m3'".split()                                                                                                                                   
Out[1]: ['m1', "'m2", "m3'"]

In [2]: import shlex                                                                                                                                           

In [3]: shlex.split("m1 'm2 m3'")                                                                                                                              
Out[3]: ['m1', 'm2 m3']
```

vruello: You're right! Perhaps the right solution would be to escape all `\` automatically before calling `shlex.split` : 
```python
     try:
        query = query.replace('\\', '\\\\')
        return [
            [neg] + pval.split(':', 1) if ':' in pval
            else [neg, pval, None]
            for neg, pval in (
                (True, x[1:]) if x[:1] in '!-' else (False, x)
                for x in shlex.split(query)
            )
        ]
```
Does it feel right to you?

p-l-: Could be. We have to think about this and check if we are not forgetting other corner cases.

vruello: I'm going to update the code. We'll see if one of the tests fails (every corner cases are not tested, but still it's a first step).

PR #675, opened 2019-03-19
Enhance utils

Nutigoodpipes:
Add tests for utils, add IPv6 address types, and fixes ip2bin.

### DIFF
--- a/ivre/utils.py
+++ b/ivre/utils.py
@@ -173,6 +173,7 @@ mapping.
         return struct.pack('!QQ', ipval >> 64, ipval & 0xffffffffffffffff)
     except TypeError:
         pass
+    raw_ipval = ipval
     try:
         ipval = ipval.decode()
     except UnicodeDecodeError:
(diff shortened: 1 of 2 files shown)


ivre/utils.py
```diff
@@ -197,7 +197,8 @@ def ip2bin(ipval):
     if len(ipval) == 16:
         return ipval
     if len(ipval) == 4:
-        return b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff' + ipval
+        return b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff' + \
+            bytes(ipval, 'utf-8')
```
p-l-: Is this necessary? I **think** ipval should already be raw bytes here (or at least this is what we assume).

tests/tests.py
```diff
@@ -2245,12 +2254,90 @@ def test_utils(self):
...
+            ivre.utils.ip2bin(b'\xe6')
+        # else case
+        with self.assertRaises(ValueError):
+            ivre.utils.ip2bin(b'23T')
+        self.assertEqual(
+            ivre.utils.ip2bin(b'23TT'),
+            b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff23TT'
+        )
+        self.assertEqual(ivre.utils.ip2bin(b'T3STTESTTESTTEST'),
+                         'T3STTESTTESTTEST')
```
p-l-:
```suggestion
                         b'T3STTESTTESTTEST')
```
(since the output is supposed to be "binary".

PR #676, opened 2019-03-19
Fix db2view

Nutigoodpipes:
Fixes db2view category option (rename to --view-category) and passive host address filter, and add tests.


ivre/tools/db2view.py
```diff
@@ -71,18 +71,43 @@ def my_parse_args():
...
                                            parents=[db.passive.argparser])
         passparser.add_argument('ips', nargs='*')
+        subparsers.add_parser('all')
 
     args = parser.parse_args()
 
-    if args.category:
-        db.view.category = args.category
+    if args.view_category:
+        db.view.category = args.view_category
```
p-l-: Didn't we conclude that this could not be a `db.view` attribute (thread / fork unsafe + weird...)? Or do I forget something?

Nutigoodpipes: This attribute is absolutely a concurrency issue. I'll get rid of this.

Nutigoodpipes: So I think passing parameter is the best way here.

p-l-:
Did you stop the tests on purpose?

PR #677, opened 2019-03-19
Enhance --explain

Nutigoodpipes:
Fixes explain option with MongoDB Nmap and view, and add this option with PostgreSQL.


ivre/db/sql/__init__.py
```diff
@@ -880,8 +889,8 @@ def get_ips(self, flt, limit=None, skip=None):
         return tuple(action(flt, limit=limit, skip=skip)
                      for action in [self.get, self.count])
 
-    def get(self, flt, limit=None, skip=None, sort=None,
-            **kargs):
+    def _get(self, flt, limit=None, skip=None, sort=None, fields=None):
+        utils.LOGGER.warning("Argument 'fields' provided but unused.")
```
p-l-:
```suggestion
        if fields is not None:
            utils.LOGGER.warning("Argument 'fields' provided but not used in this backend.")
```

tests/tests.py
```diff
@@ -2652,10 +2676,12 @@ def test_50_view(self):
...
         ret, out, _ = RUN(["ivre", "db2view", "--test", "passive"])
         self.assertEqual(ret, 0)
-        self.check_value("view_test_passive", len(out.splitlines()))
+        # One entry in test should actually be one entry at the end.
+        self.check_value("view_count_passive", len(out.splitlines()))
         ret, out, _ = RUN(["ivre", "db2view", "--test", "nmap"])
         self.assertEqual(ret, 0)
-        self.check_value("view_test_active", len(out.splitlines()))
+        # One entry in test should actually be one entry at the end.
+        self.check_value("view_count_active", len(out.splitlines()))
```
p-l-: Why were those tests broken?

p-l-:
We have to check why codacy is complaining about argument mismatch.

p-l-:
The broken tests with PostgreSQL & Python 3.3 should be fixed in #684.

PR #699, opened 2019-06-08
DB: fix MetaDB() and *DB*() init

p-l-:
This will speed up tools (CLI, Web app) startup and ivre.db import.

### DIFF
--- a/ivre/db/__init__.py
+++ b/ivre/db/__init__.py
@@ -2405,79 +2405,124 @@ class DBFlow(DB):
 
 class MetaDB(object):
 
+    # Backend-specific purpose-specific sub-classes (e.g.,
+    # MongoDBNmap) must be "registered" in this dict.
+    #
+    # The keys are the purposes ("nmap", "view", "passive", etc.), and
+    # the values are dict objects which, in turn, associate a backend
+    # name (as used as the scheme of the URLs in the configuration DB*
+    # values, such as "mongodb", "http", "postgresql", etc.) to
+    # tuples; the first element of those tuples is the sub-module name
+    # ("mongo" for "ivre.db.mongo"), and the second is the class name.
+    #
+    # {"purpose": {"scheme": ("module_name", "ClassName"),
+    #              [...]},
+    #  [...]}
     db_types = {
-        "nmap": {},
-        "passive": {},
-        "data": {},
-        "agent": {},
-        "flow": {},
-        "view": {},
+        "nmap": {"http": ("http", "HttpDBNmap"),
+                 "mongodb": ("mongo", "MongoDBNmap"),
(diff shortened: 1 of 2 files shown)


ivre/db/__init__.py
```diff
@@ -2406,78 +2406,108 @@ class DBFlow(DB):
 class MetaDB(object):
 
     db_types = {
-        "nmap": {},
-        "passive": {},
-        "data": {},
-        "agent": {},
-        "flow": {},
-        "view": {},
+        "nmap": {"http": ("http", "HttpDBNmap"),
```
vruello: Maybe it would be a good idea to add a few comments to explain the structure of this dictionary.

ivre/db/__init__.py
```diff
@@ -2406,78 +2406,108 @@ class DBFlow(DB):
...
+            except (KeyError, TypeError):
+                utils.LOGGER.error(
+                    'Cannot get database for %s from %s',
+                    purpose,
+                    url.geturl(),
+                    exc_info=True,
+                )
+                return None
+            try:
+                module = __import__('ivre.db.%s' % modulename).db
```
vruello: Shouldn't it be `module = __import__('ivre.db.%s' % modulename)` (without `.db`) ?

ivre/db/__init__.py
```diff
@@ -2406,78 +2406,108 @@ class DBFlow(DB):
...
+            except ImportError:
+                utils.LOGGER.error(
+                    'Cannot import ivre.db.%s for %s',
+                    modulename,
+                    url.geturl(),
+                    exc_info=True,
+                )
+                return None
+            for submod in modulename.split('.'):
```
vruello: I don't understand why this is necessary, but it is probably linked to line 2497. If `module` is a module object representing `ivre.db.sql.sqlite`, then line 2508 should work directly.

ivre/db/__init__.py
p-l-: Nope; `__import__('ivre.db.mongo')` returns the `ivre` module. You have to access its `db` attribute:
```Python
__import__('ivre.db.mongo').db.mongo.MongoDBNmap                                                                                                       
# ivre.db.mongo.MongoDBNmap
__import__('ivre.db.mongo').MongoDBNmap
# Traceback (most recent call last):
#   File "<stdin>", line 1, in <module>
# AttributeError: module 'ivre' has no attribute 'MongoDBNmap'
```

That being said, while reading the doc to make sure I was answering your comment correctly, I saw I should probably use `importlib.import_module`. It is recommended and it would avoid that part of the code.

ivre/db/__init__.py
p-l-: Yep, that's the same question (and the same answer).

PR #702, opened 2019-06-18
Handled stringaux and stdnse

CyrilleFranchet:
Since NMap 7.70 string operations have been moved to stringaux.lua. This PR checks if stringaux is present, if not the old stdnse is used.


p-l-:
Thanks!

PR #724, opened 2019-07-09
 Nmap/XML: fix structured output for fcrdns and rpcinfo scripts

p-l-:
Those scripts use data (hostnames for fcrdns, program numbers for rpcinfo) as keys in their structured output. This is undesirable in our databases.

This PR also fixes bugs spotted in schema migrations.


ivre/db/sql/__init__.py
```diff
@@ -809,13 +811,82 @@ def __migrate_schema_9_10(self):
...
         """
         raise NotImplementedError
 
+    def _migrate_schema_11_12(self):
+        """Converts a record from version 11 to version 12. Version 12 changes
+the structured output for fcrdns and rpcinfo script.
+
+        """
+        failed = set()
+        req = (select([self.tables.scan.id,
```
vruello: You could maybe merge the two `select` requests and avoid some code duplication.

```Python
req = (select([self.tables.scan.id,
    self.tables.script.name,
    self.tables.script.port,
    self.tables.script.output,
    self.tables.script.data])
.select_from(join(join(self.tables.scan, self.tables.port),
    self.tables.script))
.where(and_(self.tables.scan.schema_version == 11,
    or_(self.tables.script.name == "fcrdns",
        self.tables.script.name == "rpcinfo")))
for rec in self.db.execute(req):
    try:
        if 'fcrdns' in rec.data:
            # Since dots in keys are converted to
            # underscores, we need to convert them
            # back. It's OK to do so because underscores
            # cannot be used in FQDNs.
            data = [
                dict(result, name=name.replace('_', '.'))
                for name, result in viewitems(rec.data["fcrdns"])
            ]
        else:
            data = xmlnmap.change_rpcinfo(rec.data["rpcinfo"])
        except Exception:
            utils.LOGGER.warning("Cannot migrate host %r", rec.id,
                exc_info=True)
            failed.add(rec.id)
        else:
            if data:
                self.db.execute(
                    update(self.tables.script)
                    .where(and_(self.tables.script.port == rec.port,
                        self.tables.script.name == rec.name))
                    .values(data={rec.name: data})
                    )
```

(I did not test the code above)

ivre/db/__init__.py
```diff
@@ -959,6 +962,33 @@ def __migrate_schema_hosts_10_11(doc):
...
+        for port in doc.get('ports', []):
+            for script in port.get('scripts', []):
+                if script['id'] == "fcrdns":
+                    if "fcrdns" in script:
+                        # Since dots in keys are converted to
+                        # underscores, we need to convert them
+                        # back. It's OK to do so because underscores
+                        # cannot be used in FQDNs.
+                        script["fcrdns"] = [
+                            dict(result, name=name.replace('_', '.'))
```
vruello: You might want to use your function `change_fcrdns`.

ivre/db/mongo.py
```diff
@@ -1287,6 +1294,38 @@ def convert(val):
...
+        updated = False
+        for port in doc.get('ports', []):
+            for script in port.get('scripts', []):
+                if script['id'] == "fcrdns":
+                    if "fcrdns" in script:
+                        # Since dots in keys are converted to
+                        # underscores, we need to convert them
+                        # back. It's OK to do so because underscores
+                        # cannot be used in FQDNs.
+                        script["fcrdns"] = [
```
vruello: You might want to use your function `change_fcrdns`.

ivre/db/sql/__init__.py
```diff
@@ -809,13 +811,82 @@ def __migrate_schema_9_10(self):
...
+               .where(and_(self.tables.scan.schema_version == 11,
+                           self.tables.script.name == "fcrdns")))
+        for rec in self.db.execute(req):
+            if 'fcrdns' in rec.data:
+                try:
+                    # Since dots in keys are converted to
+                    # underscores, we need to convert them
+                    # back. It's OK to do so because underscores
+                    # cannot be used in FQDNs.
+                    data = [
```
vruello: You might want to use your function `change_fcrdns`.

ivre/db/mongo.py
p-l-: `change_fcrdns()` does not replace back `_` by `.`.

ivre/db/sql/__init__.py
p-l-: Same.

ivre/db/__init__.py
p-l-: `change_fcrdns` does not replace `_` by `.`.

ivre/db/__init__.py
```diff
@@ -892,7 +893,9 @@ def __migrate_schema_hosts_9_10(doc):
             for script in port.get('scripts', []):
                 if script['id'] == "s7-info":
                     if 's7-info' in script:
-                        xmlnmap.change_s7_info_keys(script['s7-info'])
+                        script['s7-info'] = xmlnmap.change_s7_info_keys(
```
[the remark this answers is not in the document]
p-l-: You are totally right!

ivre/db/__init__.py
vruello: I missed that point. Nonetheless, maybe you should create another function rather than copy/paste the same logic in 3 different files ?

ivre/db/sql/__init__.py
p-l-: I have used `.in_()` rather than `or_()`

ivre/db/__init__.py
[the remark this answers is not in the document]
p-l-: I have removed this commit.

ivre/db/sql/__init__.py
p-l-: I have used `.in_()` instead of `or_()` but that's the same idea.

ivre/db/__init__.py
p-l-: Done!

PR #734, opened 2019-07-22
Update doc & docker after MongoDB backend for flows

p-l-:


p-l-:
Thanks @vruello, I have committed a modified version of your patch. What do you think?

vruello:
Sounds good to me.

PR #736, opened 2019-07-26
Nmap/XML: fix structured output for ms-sql-info and smb-enum-shares scripts

BAngelique:
Those scripts use data (instances for ms-sql-info, share names for smb-enum-shares) as keys in their structured output. This is undesirable in our databases.


ivre/xmlnmap.py
```diff
@@ -665,13 +666,75 @@ def change_rpcinfo(table):
...
+    for iname in list(table):
+        idata = table[iname]
+        if not isinstance(idata, dict):
+            continue
+        instances.append(dict(idata, Instance=iname.replace('_', '.')))
+        del table[iname]
+    return instances
+
+
+def change_smb_enum_shares_migrate(table):
```
p-l-: Do you think you could reuse the code from the existing `change_smb_enum_shares()` function? Or is there a reason not to?

ivre/xmlnmap.py
```diff
@@ -665,13 +666,75 @@ def change_rpcinfo(table):
             for proto, result in viewitems(protores)]
 
 
+def change_ms_sql_info(table):
+    """Adapt structured output generated by the "ms-sql-info" Nmap script. The
+structured output uses instances (hence, data) as keys, which is undesirable in
```
p-l-:
```suggestion
structured output uses instances (hence, data) as keys, which is undesirable in
```
Be careful with trailing whitespaces.

ivre/xmlnmap.py
```diff
@@ -595,6 +595,7 @@ def change_smb_enum_shares(table):
     result["shares"] = [
         dict(value, Share=key) for key, value in viewitems(table)
     ]
+    result["Shares"] = xmlnmap.change_smb_enum_shares_migrate(table)
```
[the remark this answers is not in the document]
BAngelique: Yes, you are right!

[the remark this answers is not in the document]
p-l-: Please only mark the conversation as "resolved" when you have committed and pushed a fix.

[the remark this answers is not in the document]
BAngelique: Ok.

ivre/xmlnmap.py
BAngelique: Yes, thank you.

ivre/xmlnmap.py
BAngelique: Yes, I am working on it.

ivre/xmlnmap.py
```diff
@@ -589,12 +597,31 @@ def change_smb_enum_shares(table):
...
+            for elt in result["shares"]
+        ]
+    else:
+        # We need to update the structured output to avoid data in field names:
+        #
+        # Old:
+        # {"\\ServerName\ShareName": {XXX}, ...}
+        # New:
+        # [{"Share": "\\ServerName\ShareName", XXX}, ...]
+        result["shares"] = [
```
p-l-: There is (at least) one problem with this: since we cannot know the order of the numeration, we cannot know the order of the resulting list. An option would be to use a `sorted()` call to fix that:
```Python
        result["shares"] = sorted(
            (
                dict(
                    value,
                    Share=_smb_enum_shares_fix_share_name(key)
                ) for key, value in viewitems(table)
            ),
            key=lambda elt: elt["Share"],
        )
```
(I have not tested this code, but I think it should work)

What do you think?

BAngelique: You are right. It works.
Thank you.

Issue #740, opened 2019-08-02
Banner search does not seem to work in webUI


trou:
### IVRE Version

```console
root@ivre:~# ivre version
IVRE - Network recon framework
Copyright 2011 - 2019 Pierre LALET <pierre.lalet@cea.fr>
Version 0.9.13

Python 3.7.3 (default, Apr  3 2019, 05:39:12) 
[GCC 8.3.0]

Linux ivre 4.19.0-5-amd64 #1 SMP Debian 4.19.37-5+deb10u1 (2019-07-19) x86_64

Dependencies:
    Python module Crypto: 2.6.1
    Python module pymongo: 3.7.1
    Python module py2neo: missing
    Python module sqlalchemy: missing
    Python module psycopg2: missing
    Python module krbV: missing
    Python module PIL: missing
    Python module MySQLdb: missing
    Python module dbus: 1.2.8
    Python module matplotlib: missing
    Python module bottle: 0.12.15


```

### Summary

After running a scan with `masscan --banner` , importing the scan with `ivre scan2db -c XXX banners.xml` and running `ivre db2view`, the scans appear in the web UI, however, filtering with `banner:` does not work (at all).

### Expected behavior
Entering a `banner:XXX` or `banner:.*` should return entries

### Actual behavior

No host is returned.

### How to reproduce

I'm using the mongo db backend.
I'll try to submit a small repro case next week if needed.

p-l-:
Hi,

As stated in [the documentation](https://doc.ivre.rocks/en/latest/usage/web-ui.html#command-specification), the `/[expression]/[flags]` should be used for regular expressions.

In your example, you probably want to try: `banner:/XXX/` or `banner:/.*/`.

Please re-open this issue if needed.

All the best,

Pierre

trou:
Indeed, I missed this part of the documentation, even if I read part of this page.

It would probably be nice to put some examples on the page to make it obvious for the casual reader, as the text is not obviously relevant when reading the command list below.

Or maybe put something like
```
banner:"full text"|/regex/
```
?

trou:
Or even better, in the tooltip of the web UI change from `banner:[string or regex]` to `banner:["full string" or /regexp/]` ?

p-l-:
I like your second proposition better (that part of the documentation looks already too dense to me). Changed in #743.

PR #743, opened 2019-08-07
Web: update doc

p-l-:
As suggested by @trou in #740, this will make it clear that regular expressions should be specified using the `/regexp/` syntax.


trou:
That's way clearer, thanks !

PR #785, opened 2019-09-24
Passive: support DNSBL for IPv6

### LINKED ISSUE #712: Support IPv6 in DNSBL answers
#652, implementing  #590, has added support for DNSBL answers in *passive*, but it does not support IPv6. This should be implemented.

tedbe:
Fixes #712.


p-l-:
Thanks! Regarding the failing tests:
- Elasticsearch fails are normal, I need to update a file.
- Pylint fail is related to an upgrade of Pylint in Travis-CI and will be handled in another PR.

PR #810, opened 2019-10-28
clean up docker-compose

FireFart:
the current docker-compose file had the wrong paths for mongodb volumes included. This PR fixes the paths and cleans up the docker-compose file and also adds a sample ivre.conf to add additional configs like the CSRF protection URLs.
It also specifies a mongodb version instead of latest otherwise we would automatically use a newer major version on a pull which will break stuff as this needs to be a manual update of the db

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -22,7 +22,6 @@ services:
     volumes:
       - ./mongodb:/data/db
       - ./mongoconfigdb:/data/configdb
-      - ./ivre.conf:/etc/ivre.conf:ro
     restart: always
     networks:
       - ivrenet
@@ -33,6 +32,8 @@ services:
     restart: always
     ports:
       - "80:80"
+    volumes:
+      - ./ivre.conf:/etc/ivre.conf:ro
     depends_on:
       - ivredb
     networks:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -16,22 +16,38 @@
 
 version: '3'
 services:
-  ivredb:
+  db:
+    container_name: ivre_db
     image: mongo
     volumes:
-      - ./var_lib_mongodb:/var/lib/mongodb
-      - ./var_log_mongodb:/var/log/mongodb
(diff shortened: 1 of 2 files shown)


p-l-:
Hi,

Thanks for your submission. I don't understand what problem(s) you are trying to solve here.

I don't think ivre should be used on the DB server. It is not even installed so I don't understand why we should add a config file.

Docker containers are supposed to use the latest version of IVRE, and use the latest version of MongoDB on purpose.

Also, the container names are fine as they are (if my memory serves well, at least one is used in the documentation).

FireFart:
The problem is that the mongo image uses another path for data files so the old volumes will not work. Also if you uprade from mongodb 4.2 to let's say 4.3 you might need to update the data inside the database before booting the new mongodb version. If we stay on latest, there will be unintended upgrade errors once a new version of mongodb is pushed out. The naming is just to have shorter and more consistent container names

FireFart:
For the conf I just noticed it was added to the wrong container, just fixed this. Thanks for pointing this out

p-l-:
The configuration file already exists in the Web container. There is a confusion, I did not remember that the ivre/db container was not used in the Dockerfile.

FireFart:
Yes it already exists but it will be mounted ro with it's contents so the uer can easily modify the config file and add other stuff to it like the domains for the anti csrf protection

p-l-:
The user can do that if need be, I don't see a reason why it should be done by default.

FireFart:
because to change it you would have to edit the container and restart it which is not the intended behaviour on docker. With the config file outside you just do a docker-compose stop and up again to get it running. Without the mount it will be more complicated

p-l-:
That is not what I had in mind. What I was thinking is: when a user needs a specific configuration, he can add the mount for the configuration file, of course (that's what I do to run the demo). But I don't understand why this should be done by default.

FireFart:
because I find it easier for people not that deep into docker, docker-compose and ivre to edit this file instead of having to edit the docker-compose configuration. Anyways, I removed the config file from the PR

PR #811, opened 2019-10-29
Docker (compose): use ivre/db instead of mongo container

p-l-:
Fixes #810

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -18,12 +18,14 @@ version: '3'
 services:
   ivredb:
     image: ivre/db
+    container_name: ivredb
     volumes:
       - ./var_lib_mongodb:/var/lib/mongodb
       - ./var_log_mongodb:/var/log/mongodb
     restart: always
   ivreweb:
     image: ivre/web
+    container_name: ivreweb
     restart: always
     ports:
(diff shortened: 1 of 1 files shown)


FireFart:
Why not use the official mongo:4.2 image (https://hub.docker.com/_/mongo)? I can't see any special configuration in here otherwise other data directories

p-l-:
I would like to use the same config for vagrant & docker-compose.

For now, MongoDB is the only backend in the official image, but that was not the case until very recently (we had Neo4j), and it will probably change in a near future (PostgreSQL and/or Elasticsearch).

p-l-:
However, I do agree it would probably be better to use the official MongoDB image, you are totally right.

I'm going to merge this PR, but I would totally consider a PR using the official MongoDB image (provided it updates both the `docker-compose.yml` and the `Vagrantfile`) and removing the ivre/db image.

BTW, I have created a commit using your name (I hope you don't mind) to add the names to the containers created with `docker-compose.yml`.

Thanks again.

p-l-:
[Travis-CI tests](https://travis-ci.org/p-l-/ivre/builds/604314442)

FireFart:
awesome thanks!

PR #823, opened 2019-11-13
passive_recon: log all tcp established connection as open port

Nutigoodpipes:


zeek/ivre/passiverecon/__load__.zeek
```diff
@@ -52,6 +52,7 @@ export {
         POP_SERVER,
         TCP_CLIENT_BANNER,
         TCP_SERVER_BANNER,
+	OPEN_PORT,
```
p-l-: You probably have a tab instead of 8 spaces here.
```suggestion
        OPEN_PORT,
```

PR #831, opened 2019-11-29
Flow: add cleanup step in MongoDB backend

vruello:
This is similar to the heuristic implemented in Neo4j backend.

To find out reversed flows, it searches for flows with a unique source port and a destination port greater than 128. Among these flows aggregated by source port, it takes only those with more than 5 different destination ports. 

To avoid reversing flows which correspond to active scans, it checks that the average size of TCP packets is more than 50 bytes (an empty TCP packet is 40 bytes long). In practice, a nmap SYN scan uses 44 to 48 bytes long TCP packets.

TODO : Put `mongo.nopcap`  on `ivre.rocks` and change download link in `.travis.yml`.


.travis.yml
```diff
@@ -148,6 +148,9 @@ before_script:
   - wget -q --no-check-certificate -O - https://ivre.rocks/data/tests/dnsbl.pcap.bz2 | bunzip2 > tests/samples/dnsbl.pcap
   ## nfcapd files from ivre.rocks
   - wget -q --no-check-certificate -O - https://ivre.rocks/data/tests/nfcapd > tests/samples/nfcapd
+  ## mongo.nopcap file from ivre.rocks
+  ## XXX
+  - wget -q --no-check-certificate -O - https://cloud.vruello.fr/s/i3MeDmGRsNiE8sN/download > tests/samples/mongo.nopcap
```
vruello: The `.nopcap` extension enables to insert this capture only when I want to. With `.pcap`, it would have been added to `self.pcap_files` and then inserted with the other pcap files (and thus it would have changed every test results in flow and passive...)

p-l-: Yep. I **think** it would be better to do that... Anyway, we really should improve the way we generate result values and have an easier way to do it.

The PCAP file is now here BTW: https://ivre.rocks/data/tests/mongo.pcap

vruello: I am not convinced that inserting every pcap at the same time is a right thing to do. 
If one day a bug is introduced in the cleanup phase, the test that will fail is the first one (i.e. total flows count). Because this test validates that the whole insertion happened well (including all the other pcaps), it might not be easy to notice that the bug is in the cleanup phase. 
If it had been separated, the test that would have failed is the one specific with cleanup, indicating immediately that there is a problem with flows cleanup.

p-l-: I think I get your point. In that case, since we do not want/need to test the entire bro/zeek insertion, would be to fetch & insert the `.log` file directly. What do you think?

vruello: I'm fine with inserting the `.log` file directly. 
Could you replace `mongo.pcap` by `mongo_conn.log` on `ivre.rocks`? You can find it here: https://cloud.vruello.fr/s/jgcXd3pqgKTJiBp/download (or build it yourself).

p-l-: Done!

vruello: I have updated the code accordingly.

PR #840, opened 2019-12-12
Update elastic.py

TenGbps:
For fixing elasticsearch.exceptions.RequestError: RequestError(400, u'illegal_argument_exception', u'Document contains at least one immense term in field="ports.scripts.output" (whose UTF8 encoding is longer than the max length 32766),

### DIFF
--- a/ivre/db/elastic.py
+++ b/ivre/db/elastic.py
@@ -94,7 +94,7 @@ class ElasticDB(DB):
                         "dynamic_templates": [
                             {"strings": {
                                 "match_mapping_type": "string",
-                                "mapping": {"type": "keyword"},
+                                "mapping": {"type": "keyword", "ignore_above": 32000},
                             }},
                         ],
                     }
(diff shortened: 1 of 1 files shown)


ivre/db/elastic.py
```diff
@@ -94,7 +94,7 @@ def init(self):
                         "dynamic_templates": [
                             {"strings": {
                                 "match_mapping_type": "string",
-                                "mapping": {"type": "keyword"},
+                                "mapping": {"type": "keyword", "ignore_above": 32000},
```
p-l-:
```suggestion
                                # prevent RequestError exceptions when
                                # one term's UTF-8 encoding is bigger
                                # than the max length 32766
                                "mapping": {"type": "keyword", "ignore_above": 32000},
```

TenGbps: prevent RequestError exceptions when a large document is stored (max length 32766)

p-l-: Can you commit the change if you approve it?

PR #876, opened 2020-02-06
Passive: add ja3 options in ipinfo

vruello:
`searchja3server` and `searchja3client` are already implemented and tested in passive backend specific classes.

Therefore, this is only an awful copy/paste of the code which handles ja3 options in view.


p-l-:
If you have a moment, it would be great to add a CLI test.

PR #896, opened 2020-02-28
Web/API: handle subnet requests in /passivedns/

Frky:


p-l-:
Hi,

Rather than this change (`l` => `c`), in ``, change the line:
```sh
pip install -U $PIP_INSTALL_OPTIONS flake8
```
To
```sh
pip install -U $PIP_INSTALL_OPTIONS 'flake8<3.8'
```
This will use previous flake8 version, so that you don't have to worry about that and we can get your change merged.

Frky:
Hi, thanks for the tips.
There seems to be other problems with my PR though. Working on it.

tests/tests.py
```diff
@@ -2318,6 +2318,51 @@ def test_40_passive(self):
...
+                                                         HTTPD_PORT))
+            udesc = urlopen(req)
+            self.assertEqual(udesc.getcode(), 200)
+            found = False
+            for line in udesc:
+                cur_rec = json.loads(line)
+                self.assertEqual(rec['addr'], cur_rec['addr'])
+                if cur_rec['rtype'] == rtype and \
+                        cur_rec['rrname'] == rec['value']:
+                    found = True
```
p-l-, line 2341: Once you have understood why some records may lack and "addr" field, I'd suggest that you break once the expected record has been found:
```suggestion
                    found = True
                    break
```

Frky: Thanks for the suggestion.

PR #912, opened 2020-03-28
XML/Masscan: keep the scripts at the port level so that we can have intel from both 139 and 445

p-l-:


p-l-:
Travis CI OK: https://travis-ci.org/github/cea-sec/ivre/builds/668069641

PR #914, opened 2020-04-01
Screenwords should be str, not bytes

p-l-:


p-l-:
Travis-CI OK: https://travis-ci.org/github/cea-sec/ivre/builds/669949755

PR #918, opened 2020-04-15
Fix Ivre on Windows

jbgalet:
Some quick fixes to allow Ivre to run on Windows:
- on Windows, os.uname() does not exists, use sys.platform (guaranteed to exist)
- a leading "/" prevent maxmind db extraction, strip it on Windows (there might be a better way to do it)
- Fix the alternate hashing method: .read return bytes therefore returns b"" and not ""

### DIFF
--- a/ivre/db/maxmind.py
+++ b/ivre/db/maxmind.py
@@ -25,6 +25,7 @@ files.
 import codecs
 from functools import reduce
 import os
+import sys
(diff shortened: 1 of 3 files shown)


p-l-:
Travis CI : https://travis-ci.org/github/cea-sec/ivre/builds/675218848

PR #998, opened 2020-10-08
Active: handle CPE values from Nmap fingerprints

p-l-:


p-l-:
https://travis-ci.org/github/cea-sec/ivre/builds/734218732

PR #999, opened 2020-10-09
Passive: Handle more values for SMB1 and store value in vector instead of string

Rhaev3n:


p-l- (approved):

p-l-:
https://travis-ci.org/github/cea-sec/ivre/builds/734294269

PR #1000, opened 2020-10-14
Passive: Parse NTLM messages in HTTP *-AUTHENTICATE headers

Rhaev3n:


ivre/passive.py
```diff
@@ -468,16 +476,29 @@ def _getinfos_ssh_hostkey(spec):
     return {'infos': info}
 
 
+def _getinfos_authentication(spec):
+    """
+    Parse value of *-AUTHENTICATE headers depending on the protocol used
+    """
+    if spec['value'][:4] == 'NTLM' and len(spec['value']) > 4:
+        spec['value'] = spec['value'][5:]
```
p-l-, line 484:
```suggestion
        spec['value'] = spec['value'][4:].strip()
```

ivre/passive.py
```diff
@@ -468,16 +476,29 @@ def _getinfos_ssh_hostkey(spec):
...
+
+    return {}
+
+
 def _getinfos_ntlm(spec):
     """
     Get infos from the NTLMSSP_CHALLENGE matching the smb-os-discovery scripts
     form Masscan and Nmap
     """
+
```
p-l-, line 495: Can you remove this blank line?

ivre/utils.py
```diff
@@ -2213,3 +2214,101 @@ def deep_sort_dict_list(elt):
...
+        res = s.decode('utf-16').encode()
+        if len(res) == ln // 2:
+            return res
+
+        LOGGER.warning("Size too small %r", s)
+        return None
+
+    except UnicodeDecodeError:
+        LOGGER.warning("Cannot decode %r", s)
+        return None
```
p-l-, line 2235: Actually, the failure when `ntlm_msg[offset:offset + ln] is too short would occur in `struct.unpack()` so the test for `len(res) == ln // 2` comes too late. Also, you don't really need `struct` here...
```suggestion
    s = ntlm_msg[offset:offset + ln]
    if len(s) < ln:
        LOGGER.warning("Data too small at offset %d [%r, size %d]",
                       offset, ntlm_msg, msg)
        return None
    try:
        return s.decode('utf-16').encode()
    except UnicodeDecodeError:
        LOGGER.warning("Cannot decode %r", s)
        return None
```

ivre/utils.py
```diff
@@ -2213,3 +2214,101 @@ def deep_sort_dict_list(elt):
...
+
+
+# The positions of `Negotiate Version` and `Negotiate Target Info`
+# in the NTLM flags
+flag_version = 0b10000000000000000000000000
+flag_targetinfo = 0b100000000000000000000000
```
p-l-, line 2241:
```suggestion
flag_version = 0x2000000
flag_targetinfo = 0x800000
```

ivre/utils.py
```diff
@@ -2213,3 +2214,101 @@ def deep_sort_dict_list(elt):
...
+
+
+def _ntlm_challenge_extract(challenge):
+    """
+    Extract host information in a NTLM_CHALLENGE message
```
p-l-, line 2246:
```suggestion
    Extract host information in an NTLM_CHALLENGE message
```

ivre/utils.py
```diff
@@ -2213,3 +2214,101 @@ def deep_sort_dict_list(elt):
...
+    """
+    flags = struct.unpack('I', challenge[20:24])[0]
+
+    # Get target name
+    lntarget, offset = struct.unpack('HH', challenge[12:14] + challenge[16:18])
```
p-l-, line 2251: Check for sufficient data size!

p-l- (changes requested):

ivre/utils.py
```diff
@@ -2213,3 +2214,110 @@ def deep_sort_dict_list(elt):
...
+            while len(challenge) >= 4 and typ <= 5:
+                typ, ln = struct.unpack('HH', challenge[0:2] + challenge[2:4])
+                if typ <= 5:
+                    value = _extract_substr(challenge, 4, ln)
+                    challenge = challenge[4 + ln:]
+                    yield value
+
+        for k, v in zip(['domain', 'name', 'domain-dns', 'name-dns',
+                         'tree-dns'], _get_targetinfo(challenge)):
+            value += ",{}:{}".format(k, encode_b64(v).decode())
```
p-l-, line 2308: Use the `typ` value to find the field name, not the order.

p-l- (approved):

p-l-:
https://travis-ci.org/github/cea-sec/ivre/builds/736298924

PR #1009, opened 2020-10-22
Enh ntlm view

Rhaev3n:


ivre/db/__init__.py
```diff
@@ -1251,6 +1252,25 @@ def __migrate_schema_hosts_17_18(doc):
...
+
+        """
+        assert doc["schema_version"] == 18
+        doc["schema_version"] = 19
+        for port in doc.get('ports', []):
+            for i, script in enumerate(port.get('scripts', [])):
+                if script['id'] == 'smb-os-discovery':
+                    smb, ntlm = xmlnmap.split_smb_os_discovery(script)
+                    port.get('scripts', [])[i] = smb
+                    port.get('scripts', []).append(ntlm)
```
p-l-, line 1270: Don't do this! ;-)
You need to update the existing script, maybe using `script.update(smb)`. You don't need `port.get('scripts', [])` because if you are there, there are indeed `scripts` in `port`.
```suggestion
            for script in port.get('scripts', []):
                if script['id'] == 'smb-os-discovery':
                    smb, ntlm = xmlnmap.split_smb_os_discovery(script)
                    script.update(smb)
                    port['scripts'].append(ntlm)
```

ivre/db/mongo.py
```diff
@@ -1549,6 +1550,24 @@ def migrate_schema_hosts_17_18(doc):
...
+        """
+        assert doc["schema_version"] == 18
+        doc["schema_version"] = 19
+        for port in doc.get('ports', []):
+            for i, script in enumerate(port.get('scripts', [])):
+                if script['id'] == 'smb-os-discovery':
+                    smb, ntlm = xmlnmap.split_smb_os_discovery(script)
+                    port.get('scripts', [])[i] = smb
+                    port.get('scripts', []).append(ntlm)
+        return doc
```
p-l-, line 1569: This will not work, as you need to return an `update` document.
```suggestion
        update = {"$set": {"schema_version": 19}}
        updated = False
        for port in doc.get('ports', []):
            for script in port.get('scripts', []):
                if script['id'] == 'smb-os-discovery':
                    smb, ntlm = xmlnmap.split_smb_os_discovery(script)
                    script.update(smb)
                    port['scripts'].append(ntlm)
                    updated = True
        if updated:
            update["$set"]["ports"] = doc['ports']
        return update
```

ivre/db/mongo.py
```diff
@@ -1549,6 +1550,28 @@ def migrate_schema_hosts_17_18(doc):
...
+        for port in doc.get('ports', []):
+            for script in port.get('scripts', []):
+                if script['id'] == 'smb-os-discovery':
+                    smb, ntlm = xmlnmap.split_smb_os_discovery(script)
+                    script.update(smb)
+                    port['scripts'].append(ntlm)
+                    updated = True
+        if updated:
+            update["$set"]["ports"] = doc['ports']
+        return doc
```
p-l-, line 1573:
```suggestion
        return update
```

ivre/db/sql/__init__.py
```diff
@@ -1142,6 +1144,57 @@ def _migrate_schema_17_18(self):
...
+                except Exception:
+                    utils.LOGGER.warning("Cannot migrate host %r", rec.id,
+                                         exc_info=True)
+                    failed.add(rec.id)
+                else:
+                    if 'masscan' in smb:
+                        data = {'smb-os-discovery': smb['smb-os-discovery'],
+                                'masscan': smb['masscan']}
+                    else:
+                        data = smb['smb-os-discovery']
```
p-l-, line 1177: This is a bit weird, or do I miss something? The data structure will be very different between data from Nmap and Masscan if I understand this well.

p-l-:
Aslo, you will have to handle index migration (and use the opportunity to create new indexes if need be.

ivre/db/sql/__init__.py
```diff
@@ -1142,6 +1144,58 @@ def _migrate_schema_17_18(self):
...
+                except Exception:
+                    utils.LOGGER.warning("Cannot migrate host %r", rec.id,
+                                         exc_info=True)
+                    failed.add(rec.id)
+                else:
+                    if 'masscan' in smb:
+                        data = {'smb-os-discovery': smb['smb-os-discovery'],
+                                'masscan': smb['masscan']}
+                    else:
+                        data = {'smb-os-discovery': smb['smb-os-discovery']}
```
p-l-, line 1177: I'm not sure; we probably need to discuss that.

p-l-:
Thanks again for this amazing work @Rhaev3n! I will need some time to test it but this is going to be merged as soon as I can.

ivre/db/mongo.py
```diff
@@ -919,6 +919,14 @@ class MongoDBActive(MongoDB, DBActive):
...
                 {"sparse": True},
             ),
+            (
+                [("ports.scripts.ntlm-info.NetBIOS_Domain", pymongo.ASCENDING)],
+                {"sparse": True},
+            ),
+            (
+                [("ports.scripts.ntlm-info.Product_Version", pymongo.ASCENDING)],
+                {"sparse": True},
+            ),
```
p-l-, line 929: I think this should be added to the `schema_migrations_indexes` attribute, so that they get created when a user migrates its data. What do you think?

p-l- (approved):
Appart from the index creation during migration, that looks good to me!

p-l-:
Something like that would do:
```diff
--- a/ivre/db/mongo.py
+++ b/ivre/db/mongo.py
@@ -1160,6 +1160,23 @@ class MongoDBActive(MongoDB, DBActive):
                     ),
                 ],
             },
+            19: {
+                "ensure": [
+                    (
+                        [("ports.scripts.ntlm-info.NetBIOS_Domain", pymongo.ASCENDING)],
+                        {"sparse": True},
+                    ),
+                    (
+                        [
+                            (
+                                "ports.scripts.ntlm-info.Product_Version",
+                                pymongo.ASCENDING,
+                            )
+                        ],
+                        {"sparse": True},
+                    ),
+                ]
+            },
         },
     ]
     schema_latest_versions = [
```

PR #1036, opened 2020-11-25
Searchntlm: NTLM topvalues and filter on NTLM scripts

Rhaev3n:


ivre/db/mongo.py
```diff
@@ -2390,6 +2390,8 @@ def searchscript(cls, name=None, output=None, values=None, neg=False):
         if name is not None:
             if isinstance(name, list):
                 req["id"] = {"$in": name}
+                # Used to find the corresponding alias if 'values' is given
+                name = name[0]
             else:
                 req["id"] = name
```
p-l-: Do you still need to accept lists here?

p-l-:
I **think** almost everything from this PR went into #1009, except for a tiny fix that I have integrated in #1082. Please let me know if I'm forgetting something.

PR #1042, opened 2020-12-03
Adding nmap --top-ports option in nmapopt.py

Amadimk:
I implemented, the --top-ports option for nmap in the nmapopt.py file, added some logical checks against the -p option.

to use it, just add in the variable NMAP_SCAN_TEMPLATES of your ivre config file the option "top_ports" as below
```bash
NMAP_SCAN_TEMPLATES = {
    "cyslab": {
        # Commented values are default values and to not need to be
        # specified:
         "nmap": "nmap",
         "pings": "n",
         "scans": "V",
         "osdetect": False,
         "traceroute": False,
        # "resolve": 1,
         "verbosity": 0,
        # "ports":'1-1000',
	 "top_ports": "100",
         "ports": "",
         "host_timeout": "15m",  # default value: None
         "script_timeout": "2m",  # default value: None
         "scripts_categories": ['default',], # 'discovery',
        #                       'auth'],  # default value: None
        # "scripts_exclude": ['broadcast', 'brute', 'dos',
        #                    'exploit', 'external', 'fuzzer',
        #                    'intrusive'],  # default value: None
        # "scripts_force": None,
        # "extra_options": ["--top-ports 100",]
    }
}
```

the command below should produce the following result : 
```bash
$ ivre runscans --network 192.168.1.0/24 --nmap-template cyslab --output CommandLine

nmap -Pn -sC -sV --top-ports 100 --host-timeout 15m --script-timeout 2m
```

Fixes #1041.

PR #1043, opened 2020-12-04
Updating config.py to support nmap option --top-ports

Amadimk:
Adding nmap option --top-ports


p-l-:
Please update #1042 instead.

PR #1068, opened 2021-01-05
Active: (early & limited) support for Nuclei (ProjectDiscovery)

p-l-:
See https://github.com/projectdiscovery/nuclei
Cc @0xtavian
ivre-nuclei


p-l-:
We probably need to see how/if projectdiscovery/nuclei#468 is fixed; also, timestamped results would be really nice!

PR #1075, opened 2021-01-12
Zeek/NTLM: Fixed typos and small issues

Rhaev3n:


zeek/ivre/passiverecon/ntlm.zeek
```diff
@@ -35,7 +35,9 @@ function _get_protocol_version(c: connection): string {
 function _get_source(c: connection): string {
     local protocols = vector();
     for (p in c$service) {
-        protocols += p;
+        if (strcmp(p, "GSSAPI") != 0) {
```
p-l-, line 38: I don't understand. Wouldn't this work?
```suggestion
        if (p != "GSSAPI") {
```

zeek/ivre/passiverecon/ntlm.zeek
```diff
@@ -32,16 +32,22 @@ function _get_protocol_version(c: connection): string {
 }
 
 # Returns a string made from the list of protocols detected by Zeek
-function _get_source(c: connection): string {
+function _get_source(c: connection, src: port, proto: string &default=""): string {
     local protocols = vector();
     for (p in c$service) {
-        protocols += p;
+        if (p !in "GSSAPI, NTLM" && p != proto) {
```
p-l-, line 38: You are testing string "in", I'd rather have `p !in {"GSSAPI", "NTLM"}` (here, when p is `"G"`, the condition does not match and the protocol is not added).

p-l- (approved):

PR #1133, opened 2021-05-28
Modified function ivre ipdata --download

Josue87:
### IVRE Version
Insert here the output of the command ivre version, displayed as code. For example:

```
$ ivre version
Copyright 2011 - 2020 Pierre LALET <pierre@droids-corp.org>
Version 0.9.16

Python 3.9.1+ (default, Feb  5 2021, 13:46:56) 
[GCC 10.2.1 20210110]

Linux kali 5.10.0-kali3-amd64 #1 SMP Debian 5.10.13-1kali1 (2021-02-08) x86_64

Dependencies:
    Python module pymongo: 3.11.4
    Python module sqlalchemy: 1.3.22
    Python module psycopg2: 2.8.6 (dt dec pq3 ext lo64)
    Python module krbV: missing
    Python module PIL: 8.1.0
    Python module MySQLdb: 1.4.4
    Python module dbus: 1.2.16
    Python module matplotlib: 3.3.4
    Python module bottle: 0.12.19
    Python module OpenSSL: 20.0.1
    Python module tinydb: 3.15.2
```

### Summary

Ivre fails to download the file iso3166.csv, returns a 404 and aborts further downloads.

### Expected behavior

The download link has been updated, and an exception control has been added, if one of the links fails it will continue downloading the rest and the user will be able to add by hand the ones that may be missing.

### How to reproduce

sudo ivre ipdata --download

### More

The error:

image

Best approach:

image

### DIFF
--- a/.travis/ivre.conf
+++ b/.travis/ivre.conf
@@ -28,7 +28,7 @@ IPDATA_URLS = {
     'https://ivre.rocks/data/tests/db/GeoLite2-ASN.tar.gz',
     'GeoLite2-ASN-CSV.zip':
     'https://ivre.rocks/data/tests/db/GeoLite2-ASN-CSV.zip',
-    'iso3166.csv': 'https://ivre.rocks/data/tests/db/iso3166.csv',
+    'iso3166.csv': 'https://dev.maxmind.com/csv-files/codes/iso3166.csv',
     'BGP.raw': 'https://ivre.rocks/data/tests/db/data-raw-table',
 }
 NMAP_SCAN_TEMPLATES["default"]["pings"] = []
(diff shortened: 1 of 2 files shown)


p-l-:
If I understand correctly, this bug has been fixed in #1128. Also, this PR does not fix the bug, it changes a value that is only used for tests in Travis-CI.

Please

Josue87:
It's not really the problem in .travis, they are simple URLs. The problem is in ivre/geoiputils using opener.open without try block, any exception breaks the function without control.  As you can see in the attached screenshots...

The url that gave 404 only showed the problem, if all of them are correct the failure does not arrive.

p-l-:
You can create another PR, without changing the URLs? Also, the warning message should be explicit, not the exception itself. Finally, if one download has failed, the exit code should not be 0.

PR #1148, opened 2021-07-14
Zeek: support v4

p-l-:


p-l-:
Tests for Zeek v4.0 are broken: see https://github.com/p-l-/ivre/runs/3068672440

PR #1178, opened 2021-09-26
Improve how scanners are displayed

gpotter2:
part of https://github.com/ivre/ivre/pull/1175

This PR:
- a few improvements to how scanners are displayed (now also displays which probe were found to be able to identify false positives, if any)

### DIFF
--- a/ivre/active/data.py
+++ b/ivre/active/data.py
@@ -498,11 +498,20 @@ def merge_scanner_scripts(
             )
         )
     if scanners:
+
+        def _fmt_sc(sc: Dict[str, Any]) -> str:
+            res = sc["name"]  # type: str
+            if "probes" in sc:
+                res += " [%s]" % ", ".join(
+                    "%s/%s" % (x["name"], x["proto"]) for x in sc["probes"]
+                )
+            return res
+
         output.append(
-            "Scanner%s: %s"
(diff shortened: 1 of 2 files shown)


p-l- (dismissed):

p-l-:
Hi, it seems that [black](https://github.com/psf/black) is not happy with your PR. Can you have a look?

p-l-:
Still KO (black again)

gpotter2:
Sorry about that I was using ubuntu's black which is outdated.

p-l-:
Now it seems you have a typing problem

p-l- (approved):

p-l- (changes requested):
Sorry I missed that but this will only affect merged hosts, so you may want to apply a similar change to [ivre/view#L253](https://github.com/ivre/ivre/blob/master/ivre/view.py#L153) so that host that do not require a merge for the pseudo script `scanner` get the new output.

gpotter2:
Makes sense. I only tested this on passive where merging always happen (https://github.com/ivre/ivre/blob/96963faf1c0e82bdcfa94e02af7963d68de8ca16/ivre/view.py#L722)

p-l- (approved):

PR #1179, opened 2021-09-26
Detect NMAP using its ports

gpotter2:
part of https://github.com/ivre/ivre/pull/1175

This PR:
- detects nmap when it uses a default list of 1000 ports


p-l-:
One question when merging records: what would happen if we merge a record with Nmap top 1k ports with another record with 1000 other (say, random) ports? Would we remove the Nmap tag?

p-l-:
That's what I understand too. Don't you think the tag should be removed in such situations?

ivre/active/data.py
```diff
@@ -417,6 +418,17 @@ def merge_scanner_scripts(
         nports = len(ports["ports"])
         res["ports"][proto]["count"] = nports
         res["ports"]["count"] = res["ports"].get("count", 0) + nports
+        # If the list of scanned ports is almost exactly the same as
+        # the list of ports scanned by default by a scanner, assume it
+        # is.
+        for (dports, scanner_name) in DEFAULT_SCANNED_PORTS.get(proto, []):
+            if scanner_name not in res.get("scanners", []):
+                if abs(nports / len(dports) - 1) < 0.03 and set(
```
gpotter2, line 426: This checks that the set of all scanned ports matches the 1000 nmap ports with an 97% accuracy. I'd say that if this ever happens, the host was indeed using nmap at some point, even if some ports are added later.
I'm not sure I understand the case, but I can of course update it

p-l-: I understand what you mean, and even if I keep thinking it would be better to remove such a label when more ports are added, I don't have a strong argument to support that, so I'm fine with your code as it is.

Thanks for the discussion!

ivre/data/scanners.py
```diff
@@ -70,3 +70,23 @@
...
+            "Nmap"
```
p-l-, line 83: I think you could add another element to the tuple, such as "top 1000 ports" here...
```suggestion
            "Nmap",
            "Top 1000 ports (default)",
```

ivre/data/scanners.py
```diff
@@ -70,3 +70,23 @@
...
+        ),
+        # Reduced (-F) scan
+        (
+            {7, 9, 13, 21, 22, 23, 25, 26, 37, 53, 79, 80, 81, 88, 106, 110, 111, 113, 119, 135, 139, 143, 144, 179, 199, 389, 427, 443, 444, 445, 465, 513, 514, 515, 543, 544, 548, 554, 587, 631, 646, 873, 990, 993, 995, 1025, 1026, 1027, 1028, 1029, 1110, 1433, 1720, 1723, 1755, 1900, 2000, 2001, 2049, 2121, 2717, 3000, 3128, 3306, 3389, 3986, 4899, 5000, 5009, 5051, 5060, 5101, 5190, 5357, 5432, 5631, 5666, 5800, 5900, 6000, 6001, 6646, 7070, 8000, 8008, 8009, 8080, 8081, 8443, 8888, 9100, 9999, 10000, 32768, 49152, 49153, 49154, 49155, 49156, 49157},
+            "Nmap"
```
p-l-, line 88: ... and "top 100 ports" here:
```suggestion
            "Nmap",
            "Top 100 ports (fast)",
```

ivre/active/data.py
```diff
@@ -417,6 +418,17 @@ def merge_scanner_scripts(
         nports = len(ports["ports"])
         res["ports"][proto]["count"] = nports
         res["ports"]["count"] = res["ports"].get("count", 0) + nports
+        # If the list of scanned ports is almost exactly the same as
+        # the list of ports scanned by default by a scanner, assume it
+        # is.
+        for (dports, scanner_name) in DEFAULT_SCANNED_PORTS.get(proto, []):
```
p-l-, line 424:
```suggestion
        for (dports, scanner_name, scanner_info) in DEFAULT_SCANNED_PORTS.get(proto, []):
```

ivre/active/data.py
```diff
@@ -417,6 +418,17 @@ def merge_scanner_scripts(
...
+            if scanner_name not in res.get("scanners", []):
+                if abs(nports / len(dports) - 1) < 0.03 and set(
+                    ports["ports"]
+                ).issubset(dports):
+                    res.setdefault("scanners", {}).setdefault(scanner_name, set()).add(
+                        (proto, "Default ports")
```
p-l-, line 430:
```suggestion
                        (proto, scanner_info)
```

p-l- (commented):
I think we could add this specification. What do you think?

p-l-:
Hi @gpotter2,

You should run `black -t py36` against your code to fix the linting failure.

For the MongoDB failures, they do not come from your code, and have been fixed in #1248 (newer version of PyMongo no longer supports APIs that were used in IVRE). Just rebase your code against current master and everything should be fine.

Thanks!

ivre/active/data.py
```diff
@@ -424,6 +425,17 @@ def merge_scanner_scripts(
         nports = len(ports["ports"])
         res["ports"][proto]["count"] = nports
         res["ports"]["count"] = res["ports"].get("count", 0) + nports
+        # If the list of scanned ports is almost exactly the same as
+        # the list of ports scanned by default by a scanner, assume it
+        # is.
+        for (dports, scanner_name) in DEFAULT_SCANNED_PORTS.get(proto, []):
+            if scanner_name not in res.get("scanners", []):
```
p-l-, line 432: Maybe you could remove this test, since you use `res.setdefault("scanners", {}).setdefault(scanner_name, set()).add(...)`

ivre/active/data.py
p-l-: Why did you remove this change? Was it on purpose or a side effect of the rebase?

p-l- (changes requested):

ivre/active/data.py
gpotter2: If I read the code correctly, that's necessary to not re-add the value to the set (in the add()) if it exists.

gpotter2:
Really sorry for the delay

PR #1190, opened 2021-10-02
Basic Port Scanner in Python

Ayush765:


p-l-:
Hi,

Thanks for your interest in IVRE. Unfortunately, we cannot accept your contribution as it does not integrate in the project at all. Also, excellent port scanners already exist, I don't think another port scanner with nothing specific, is really needed.

Moreover, this code seems to have been copied from there, so you are probably not the author: <https://www.geeksforgeeks.org/port-scanner-using-python/>.

If you want to contribute to IVRE, please have a look at the existing issues to see where you could start.

Based on your profile, you seem interested in Hacktoberfest, which is great. Please read their helpful resources on how to get started! https://hacktoberfest.digitalocean.com/resources

PR #1192, opened 2021-10-06
Added a filter for honeypots.

huynguy97:
This is a filter that shows hosts that have been labelled as a honeypot. Only works for MongoDB databases and no tests have been made for this.

### DIFF
--- a/doc/usage/web-ui.rst
+++ b/doc/usage/web-ui.rst
@@ -138,6 +138,7 @@ Filters
 
 - ``[!]host:[IP address]`` filter a specific IP address. Using the IP
   address directly (without ``host:``) is equivalent.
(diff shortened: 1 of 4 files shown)


p-l- (changes requested):
Thanks for this PR. I'd like to suggest a change.

Also, as mentioned on the chat, we need an implementation for every other backend (SQL, Elastic, TinyDB), and tests (for that part, we will probably need to add a sample so that can be done later).

Thanks again!

p-l-:
Hi, since we now use tags, this PR is obsolete.

A simple `tag:Honeypot` filter will work (or `tag:Honeypot:"SYN+ACK honeypot"` to be more specific).

Thanks a lot!

PR #1197, opened 2021-10-13
Add STUN to zeek passiverecon script

Frky:
This adds the detection of STUN requests over UDP based on two regex : 

2021-10-13-172530_666x69_scrot

**Note**: it requires to enable inspection for every incomping UDP packet with a payload ; maybe this is not something we want because of overhead?
```
redef udp_content_deliver_all_orig = T;
```
A compromise could be to enable only for known STUN ports (*e.g.*, 3478).

### DIFF
--- a/ivre/passive.py
+++ b/ivre/passive.py
@@ -355,6 +355,8 @@ def _prepare_rec(spec, ignorenets, neverignore):
                             "service_extrainfo": "UDP payload %s" % payload,
                         }
                     )
+    elif spec["recontype"] == "STUN_HONEYPOT_REQUEST":
+        spec["value"] = utils.nmap_decode_data(spec["value"])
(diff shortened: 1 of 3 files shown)


Frky:
From what I observe, there are common values that are used frequently:
- [magic] + `000000000000000000000000`
- [magic] + `3f50d62de94fbd361b7b66d7`
- `12233445566778899001122334455667`

I haven't done yet the research to know what tool use each of those, but it seems likely that there are fixed values begin used.

p-l-:
Do you observe several values for `[magic]`?

Frky:
No, the magic is always `0x2112A442` according to [RFC5389](https://datatracker.ietf.org/doc/html/rfc5389).
If different, then the four bytes are included in the TID (and not considered as magic).

PR #1224, opened 2021-11-17
Docker/client: install p0f

### LINKED ISSUE #1223: Install p0f in Docker image ivre/client
Since p0f is supported again (#1169), we should install it in the client Docker image.

p-l-:
Fixes #1223


p-l-:
Fixes #1223

PR #1405, opened 2022-07-17
Add filter in Explore view for script:xxx

### LINKED ISSUE #1404: Explore view filter when clicking on title does not work for script:xxxx
Here I come again ;)

### IVRE Version

```
Copyright 2011 - 2022 Pierre LALET <pierre@droids-corp.org>
Version 0.9.18.dev25-dockerhub

Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110]

Linux 2f86dcf6bc91 4.9.0-19-amd64 #1 SMP Debian 4.9.320-2 (2022-06-30) x86_64

Dependencies:
    Python module pymongo: 3.11.0
    Python module sqlalchemy: missing
    Python module psycopg2: missing
    Python module krbV: missing
    Python module pycurl: missing
    Python module PIL: 8.1.2
    Python module MySQLdb: missing
    Python module dbus: missing
    Python module matplotlib: missing
    Python module bottle: 0.12.19
    Python module OpenSSL: 20.0.1
    Python module tinydb: missing

```

### Summary

Can't click on a line from the `script:http-nuclei` Explore view to filter records in the list view

### Expected behavior

Should display filtered resulting according to the line clicked, like, for example, in `script` Explore view

### Actual behavior

Does not work

### How to reproduce

In the Explore box, and with some records from nuclei in the db, enter `script:http-nuclei` and press enter.
Click on a line in the displayed graph

### More

I have tried to activate the click by creating the `script:` condition in the graphjs file at this position
https://github.com/ivre/ivre/blob/ac970494a51b00b6e0c0a570ab5c814ec5056a41/web/static/ivre/graph.js#L271

It works, but if the text contains a line feed, it fails to filter the list
Also when it works, and if I have multiple pages, it fails when loading the 2nd page because it removes the escape character.

psyray:
Fix #1404

### DIFF
--- a/web/static/ivre/graph.js
+++ b/web/static/ivre/graph.js
@@ -94,6 +94,10 @@ var GraphTopValues = (function(_super) {
 	this.filename = "TopValues";
     }
 
+    function escapeRegex(string) {
+        return '\\"'+string.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\\\$&').trim()+'\\"';
+    }
+
     $.extend(GraphTopValues.prototype, _super.prototype, {
 	get_url: function() {
 	    return 'cgi/view/top/' + encodeURIComponent(this.field) + ':' +
@@ -268,6 +272,11 @@ var GraphTopValues = (function(_super) {
 		    return 'setparam(FILTER, "script", "' + x + '");';
 		};
 	    }
+            else if(field.substr(0, 7) === 'script:') {
+                preparefilter = function(x) {
+                    return 'setparam(FILTER, "'+ field +'", "/' + escapeRegex(x) + '/");';
+                };
+            }
 	    else if(field === 'port') {
 		prepareoutput = function(x) {
 		    return x.join(' / ');
@@ -1305,3 +1314,4 @@ var GraphDiffCategories = (function(_super) {
 
     return GraphDiffCategories;
 })(GraphIpPort);
+


web/static/ivre/graph.js
```diff
@@ -94,6 +94,10 @@ var GraphTopValues = (function(_super) {
 	this.filename = "TopValues";
     }
 
+    function escapeRegex(string) {
+        return '\\"'+string.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\\\$&').trim()+'\\"';
+    }
```
p-l-, line 99: I'd rather remove the double quotes added here as they are not part of the regexp escape process:
```suggestion
// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#escaping
function escapeRegExp(string) {
  return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
}
```

web/static/ivre/graph.js
```diff
@@ -268,6 +272,11 @@ var GraphTopValues = (function(_super) {
 		    return 'setparam(FILTER, "script", "' + x + '");';
 		};
 	    }
+            else if(field.substr(0, 7) === 'script:') {
+                preparefilter = function(x) {
+                    return 'setparam(FILTER, "'+ field +'", "/' + escapeRegex(x) + '/");setparam(FILTER, "display:'+ field +'");';
```
p-l-, line 277: I don't think we should have the `display:` directive.
```suggestion
                    return 'setparam(FILTER, "'+ field +'", "/' + escapeRegExp(x) + '/");
```

p-l- (commented):
I don't have time to test right now but I will do asap

web/static/ivre/graph.js
psyray: The double quote is here to take the string in it's whole, and simplify the condition and make more readable, and also to prevent repeat the double quote everytime.
Maybe we should name it formatRegEx instead of escapeRegEx, and set a parameter to include the double quote or not ?
But you remove trim, and it's necessary to remove the line feed, if not the Javascript will fail and drop an error "unescaped line feed"

psyray:
> I don't have time to test right now but I will do asap

No problem ;)

web/static/ivre/graph.js
p-l-: I understand but that's not what exists in IVRE so far. For now, clicking in top values creates a filter for that value, it does not limit the display. For now, I'd rather keep it that way.

web/static/ivre/graph.js
p-l-: Good point for the `.trim()`. I think it is better to add it after the call, so I did it (see below)

web/static/ivre/graph.js
psyray: > I understand but that's not what exists in IVRE so far. For now, clicking in top values creates a filter for that value, it does not limit the display. For now, I'd rather keep it that way.

No problem, you're the boss, boss ;)

Issue #1398, opened 2022-07-15
Content hidden by CSS max-height rule in host list


psyray:
### IVRE Version

```
$ ivre version
IVRE - Network recon framework
Copyright 2011 - 2022 Pierre LALET <pierre@droids-corp.org>
Version 0.9.18.dev25-dockerhub

Python 3.9.2 (default, Feb 28 2021, 17:03:44)
[GCC 10.2.1 20210110]

Linux 2f86dcf6bc91 4.9.0-19-amd64 #1 SMP Debian 4.9.320-2 (2022-06-30) x86_64

Dependencies:
    Python module pymongo: 3.11.0
    Python module sqlalchemy: missing
    Python module psycopg2: missing
    Python module cryptography: 3.3.2
    Python module krbV: missing
    Python module pycurl: missing
    Python module PIL: 8.1.2
    Python module MySQLdb: missing
    Python module dbus: missing
    Python module matplotlib: missing
    Python module bottle: 0.12.19
    Python module OpenSSL: 20.0.1
    Python module tinydb: missing

```

### Summary

Due to a CSS max-height value, content is hidden in the host list

### Expected behavior

See full content in the box

### Actual behavior

Content hidden

### How to reproduce

List host with a lot of content

### More

With max-height
![image](https://user-images.githubusercontent.com/1230954/179211288-8c78d2bf-406d-4440-98ef-1adff877a61c.png)

Without
![image](https://user-images.githubusercontent.com/1230954/179211197-736e4448-39c2-449d-bd08-1a2d7eff27d4.png)

p-l-:
This is meant to avoid a "summary" that would use too much space. You can see that I am far from being a frontend dev :)

Feel free to submit a PR if you feel like it should change.

psyray:
I understand ;)
And I think this could be enhanced, I will have a look and submit a PR ;)
And I will also enhanced the display for the Explore View.
If label is too large we can't see them
![image](https://user-images.githubusercontent.com/1230954/179405472-351c8bb1-9b1c-4dd7-a77d-e069a29c1245.png)

psyray:
> This is meant to avoid a "summary" that would use too much space. You can see that I am far from being a frontend dev :)
> 
> Feel free to submit a PR if you feel like it should change.

Instead of putting an `overflow: hidden`, I could pout an `overflow: scroll` in the ivre.css, line 142
[...]

PR #1407, opened 2022-07-20
Revert "WebUI - Prevent content to be hidden in preview mode"

p-l-:
Reverts ivre/ivre#1401
My mistake, I had not tested that with Chrome / Chromium and the result is clearly less good than the previous one.
Sorry @psyray, I hope we'll find another way to fix / handle #1398.

### DIFF
--- a/web/static/ivre/ivre.css
+++ b/web/static/ivre/ivre.css
@@ -139,7 +139,7 @@ form {
 }
 
 .preview {
-    overflow: scroll;
+    overflow: hidden;
     border: 1px solid rgba(0, 0, 0, 0.1);
     max-height: 150px;
 }


psyray:
Sorry me too, I haven't tested with Chrome.
But we can set `auto` instead of forcing `scroll`, so scrollbar are added only when necessary.
You want another PR or youd could do it yourself ?

p-l-:
> Sorry me too, I haven't tested with Chrome.

No worries!

> But we can set `auto` instead of forcing `scroll`, so scrollbar are added only when necessary. And the result is like in FF

Tested, I like it.

> Try it in the Chrome Inspect tool ;) You want another PR or you could do it yourself ?

You can do the PR, since you have done the job!

PR #1445, opened 2022-10-19
db2view: support --parallel COUNT

p-l-:
Thanks @0xDeva for that excellent idea!


p-l-:
OK, tuns out, result is slower on real-world examples!

PR #1446, opened 2022-10-21
Fix bug in zgrabout.py

Frky:
The current script fails at L413 if, for some reason, the variable `banner` is never set by the previous block. When doing `scan2db`, this bug leads to the result not being imported.

### DIFF
--- a/ivre/zgrabout.py
+++ b/ivre/zgrabout.py
@@ -300,7 +300,7 @@ def zgrap_parser_http(
     # Since Zgrab does not preserve the order of the headers, we need
     # to reconstruct a banner to use Nmap fingerprints
     # define empty banner for log in case banner is never set in the following blocks
-    banner = b""
+    banner = None
     if resp.get("headers"):
         headers = resp["headers"]
         # Check the Authenticate header first: if we requested it with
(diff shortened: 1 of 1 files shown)


p-l- (commented):
I'd rather have `banner` set to `None`, and then a test to only call `utils.match_nmap_svc_fp()` when `banner is not None`.
What do you think?

Frky:
Ok, I changed it as suggested. Is it what you meant?

ivre/zgrabout.py
```diff
@@ -408,8 +410,10 @@ def zgrap_parser_http(
                 script["masscan"] = {"raw": utils.encode_b64(banner).decode()}
             res.setdefault("scripts", []).append(script)
             handle_http_headers(hostrec, res, http_hdrs, path=url.get("path"))
-    info: NmapServiceMatch = utils.match_nmap_svc_fp(
-        banner, proto="tcp", probe="GetRequest"
+    info: NmapServiceMatch = (
+        utils.match_nmap_svc_fp(banner, proto="tcp", probe="GetRequest")
+        if banner
+        else None
```
p-l-, line 416:
```suggestion
    if banner is None:
        info: NmapServiceMatch = {}
    else:
        info: NmapServiceMatch = utils.match_nmap_svc_fp(banner, proto="tcp", probe="GetRequest")
```

p-l- (approved):

PR #1544, opened 2023-06-23
Fix bug when ZDNS answer doesn't have expected fields

KuxODdyA:


p-l-:
Hey there!

Thanks for this contribution. You should run `black` (and other linters) against your code!

PR #1545, opened 2023-06-23
Handle some corner cases to avoid dropping the whole file being imported

KuxODdyA:


ivre/utils.py
```diff
@@ -2112,7 +2119,11 @@ def get_cert_info(cert: bytes) -> Dict[str, Any]:
             hashtype: hashlib.new(hashtype, cert).hexdigest()
             for hashtype in ["md5", "sha1", "sha256"]
         }
-        data = osslc.load_certificate(osslc.FILETYPE_ASN1, cert)
+        try:
+            data = osslc.load_certificate(osslc.FILETYPE_ASN1, cert)
+        except OpenSSL.crypto.Error:
+            LOGGER.warning("Cannot decode asn1 - not enough data")
```
p-l-, line 2125: It could be another error, or am I missing something?
What about:
```suggestion
            LOGGER.warning("Cannot load certificate", exc_info=True)
```

ivre/utils.py
```diff
@@ -2112,7 +2119,11 @@ def get_cert_info(cert: bytes) -> Dict[str, Any]:
...
+            return {}
```
p-l-, line 2126: We have already computed the hashes, what about keeping them?
```suggestion
            return result
```

ivre/utils.py
```diff
@@ -2152,7 +2163,11 @@ def get_cert_info(cert: bytes) -> Dict[str, Any]:
             # RSA
             numbers = pubkey.to_cryptography_key().public_numbers()
             result["pubkey"]["exponent"] = numbers.e
-            result["pubkey"]["modulus"] = str(numbers.n)
+            try:
+                result["pubkey"]["modulus"] = str(numbers.n)
+            except ValueError:
+                sys.set_int_max_str_digits(int(math.log(numbers.n, 10) + 1))
+                result["pubkey"]["modulus"] = str(numbers.n)
```
p-l-, line 2170: Same:
```suggestion
            except ValueError as exc:
                try:
                    orig = sys.set_int_max_str_digits()
                except AttributeError:
                    raise exc
                sys.set_int_max_str_digits(int(math.log(numbers.n, 10) + 1))
                result["pubkey"]["modulus"] = str(numbers.n)
                sys.set_int_max_str_digits(orig)
```

or:
```suggestion
            except ValueError as exc:
                try:
                    sys.set_int_max_str_digits(int(math.log(numbers.n, 10) + 1))
                except AttributeError:
                    raise exc
                result["pubkey"]["modulus"] = str(numbers.n)
```

ivre/utils.py
```diff
@@ -1859,7 +1861,12 @@ def parse_ssh_key(data: bytes) -> Dict[str, Any]:
             # convert integer to strings to prevent overflow errors
             # (e.g., "MongoDB can only handle up to 8-byte ints")
             for val in ["exponent", "modulus"]:
-                info[val] = str(info[val])
+                try:
+                    # with large SSH keys, it could lead to an exception
+                    info[val] = str(info[val])
+                except ValueError:
+                    sys.set_int_max_str_digits(int(math.log(info[val], 10) + 1))
+                    info[val] = str(info[val])
```
p-l-, line 1869: Since this is new in Python 3.11, and since we might want to preserve the original value (I'm not sure on this part, I have no idea of the consequences of calling `sys.set_int_max_str_digits()`), we could do:
```suggestion
                except ValueError as exc:
                    try:
                        orig = sys.set_int_max_str_digits()
                    except AttributeError:
                        raise exc
                    sys.set_int_max_str_digits(int(math.log(info[val], 10) + 1))
                    info[val] = str(info[val])
                    sys.set_int_max_str_digits(orig)                    
```
or:
```suggestion
                except ValueError as exc:
                    try:
                        sys.set_int_max_str_digits(int(math.log(info[val], 10) + 1))
                    except AttributeError:
                        raise exc
                    info[val] = str(info[val])
```
WDYT?

p-l- (changes requested):

ivre/utils.py
```diff
@@ -38,11 +38,13 @@
 import logging
 import math
 import os
+import OpenSSL
```
p-l-, line 41: This one is not a standard Python import so it has to be imported below.

ivre/utils.py
```diff
@@ -2112,7 +2119,11 @@ def get_cert_info(cert: bytes) -> Dict[str, Any]:
             hashtype: hashlib.new(hashtype, cert).hexdigest()
             for hashtype in ["md5", "sha1", "sha256"]
         }
-        data = osslc.load_certificate(osslc.FILETYPE_ASN1, cert)
+        try:
+            data = osslc.load_certificate(osslc.FILETYPE_ASN1, cert)
+        except OpenSSL.crypto.Error:
```
p-l-, line 2124: Already imported as:
```suggestion
        except osslc.Error:
```

ivre/utils.py
p-l-, line 41: Not needed (see below)
```suggestion
```

p-l- (changes requested):
OpenSSL is not mandatory, and it was already imported. This should fix the failing tests.

ivre/utils.py
```diff
@@ -1859,7 +1860,15 @@ def parse_ssh_key(data: bytes) -> Dict[str, Any]:
...
             for val in ["exponent", "modulus"]:
-                info[val] = str(info[val])
+                try:
+                    # with large SSH keys, it could lead to an exception
+                    info[val] = str(info[val])
+                except ValueError as exc:
+                    try:
+                        sys.set_int_max_str_digits(0)
+                    except AttributeError:
+                        raise exc
```
p-l-, line 1870:
```suggestion
                        raise exc from exc
```

ivre/utils.py
```diff
@@ -2152,7 +2165,14 @@ def get_cert_info(cert: bytes) -> Dict[str, Any]:
...
             numbers = pubkey.to_cryptography_key().public_numbers()
             result["pubkey"]["exponent"] = numbers.e
-            result["pubkey"]["modulus"] = str(numbers.n)
+            try:
+                result["pubkey"]["modulus"] = str(numbers.n)
+            except ValueError as exc:
+                try:
+                    sys.set_int_max_str_digits(0)
+                except AttributeError:
+                    raise exc
```
p-l-, line 2174:
```suggestion
                    raise exc from exc
```

p-l- (changes requested):

p-l- (approved):

PR #1684, opened 2024-12-20
add env variable to set docker mount point

exzz:
Hello Pierre,

Here is a small modification to change the default docker compose mount configuration.
I would like to be able to set `IVRE_MONGODB_PATH`, `IVRE_DOKUWIKI_PATH` and `IVRE_SHARE_PATH` to change default docker host mount point.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -20,7 +20,7 @@ services:
     image: mongo
     container_name: ivredb
     volumes:
-      - ./var_lib_mongodb:/data/db
+      - ${IVRE_MONGODB_PATH:-/var_lib_mongodb}:/data/db
     restart: always
   ivreuwsgi:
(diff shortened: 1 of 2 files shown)


docker/docker-compose.yml
```diff
@@ -20,7 +20,7 @@ services:
     image: mongo
     container_name: ivredb
     volumes:
-      - ./var_lib_mongodb:/data/db
+      - ${IVRE_MONGODB_PATH:-/var_lib_mongodb}:/data/db
```
p-l-, line 23:
```suggestion
      - ${IVRE_MONGODB_PATH:-./var_lib_mongodb}:/data/db
```

p-l- (commented):
Hi,

Thanks for your submission. Did you test your PR, is it working as you need?

PR #1739, opened 2025-11-19
Handle invalid port values in CLIs

vratiskol:
- Add utils.InvalidPort and a shared parse_port() helper (enforces 1–65535 range) and have the DB argument parser call it for every option that accepts a port.
  - Update the CLI entry points (ipinfo, scancli, view) to catch InvalidPort and report a clean error, just like we already do for invalid IPs.
  
Goal: prevents tracebacks when users pass a fancy string or out-of-range number via --port.


p-l- (approved):

p-l-:
Thanks a lot for this contribution!

PR #1741, opened 2025-11-21
CLI - Handle Ctrl+C exit cleanly

vratiskol:
**Ctrl+C handling improvements:**  

- Ensure ivre CLI exits cleanly on Ctrl+C by wrapping display pagination in a KeyboardInterrupt handler and returning exit code 130 without tracebacks.

- Improve pagination robustness by handling Ctrl+D alongside Ctrl+C and always emitting a trailing newline when interrupted.

### DIFF
--- a/ivre/nmapout.py
+++ b/ivre/nmapout.py
@@ -217,10 +217,6 @@ def displayhosts(recordsgen, out=sys.stdout, **kargs):
                     # Gracefully exit pagination on Ctrl+D
                     out.write("\n")
                     return
-                except KeyboardInterrupt:
-                    # Tidy output, then exit with the conventional SIGINT code
-                    out.write("\n")
-                    raise SystemExit(130) from None
             else:
                 out.write("\n")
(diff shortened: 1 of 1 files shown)


ivre/nmapout.py
```diff
@@ -207,12 +207,26 @@ def displayhosts(recordsgen, out=sys.stdout, **kargs):
...
+                try:
+                    input()
+                except EOFError:
+                    # Gracefully exit pagination on Ctrl+D
+                    out.write("\n")
+                    return
+                except KeyboardInterrupt:
+                    # Tidy output, then exit with the conventional SIGINT code
+                    out.write("\n")
+                    raise SystemExit(130) from None
```
p-l-: I don't think this is needed since the outer try-except block will catch this exception
```suggestion
```

p-l- (commented):
Sorry for being picky, I'm trying to keep the code as simple as possible. Could you tell me if you agree this can be removed?

PR #1742, opened 2025-11-21
[Mongodb] Fix Mongo schema migrations warning to use session with no_cursor_timeout

vratiskol:
### Change:  

- Removes the PyMongo warning by aligning with driver expectations: servers can close non-session cursors after 30 minutes even with no_cursor_timeout=True.
  - Keeps long-running migration cursors alive more reliably, reducing the risk of mid-run CursorNotFound from idle session timeouts on the server.
  - Ensures the read cursor and its subsequent update/delete operations share the same server session, improving consistency and lowering the chance of partial progress if the server prunes unused cursors.
  - Provides a clean fallback: if the server doesn’t support sessions (e.g., DocumentDB or older setups), it transparently degrades to the previous behavior without breaking.

**ivre view --update-schema**
`/usr/local/lib64/python3.11/site-packages/pymongo/synchronous/collection.py:1920: UserWarning: use an explicit session with no_cursor_timeout=True otherwise the cursor may still timeout after 30 minutes, for more info see https://mongodb.com/docs/v4.4/reference/
  method/cursor.noCursorTimeout/#session-idle-timeout-overrides-nocursortimeout
    return Cursor(self, *args, **kwargs)`

### Tech details:

  - ivre/db/mongo.py: start a client session (fallback to nullcontext on failure), wrap the migration cursor in that context, and pass the session to both the find() and _migrate_update_record() calls so long-running migrations keep their cursor alive.
  - ivre/db/mongo.py: allow _migrate_update_record() to accept an optional session and propagate it to update/delete operations; import nullcontext for the fallback path.

### DIFF
--- a/ivre/active/data.py
+++ b/ivre/active/data.py
@@ -462,13 +462,6 @@ def merge_scanner_scripts(
     return curscript
 
 
-def format_nuclei_output(nuclei: dict[str, Any]) -> str:
-    output = "[%(severity)s] %(name)s found at %(url)s" % nuclei
-    if "favicon-hash" in nuclei:
-        output += " (favicon hash: %s)" % nuclei["favicon-hash"]
-    return output
-
-
 def merge_nuclei_scripts(
     curscript: NmapScript, script: NmapScript, script_id: str
 ) -> NmapScript:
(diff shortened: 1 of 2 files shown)


p-l-:
I think Copilot makes a valid point here, but I'm not sure how hard it would be to implement that. I'll have a look.

PR #1745, opened 2025-11-23
[Nuclei] Add nuclei favicon hash detection and formatted output

vratiskol:
**First step**

* Introduce a shared Nuclei output formatter to append favicon hashes when available.
* Parse `extracted-results` from the `favicon-detect` template (handling both hyphen and underscore keys), store the mmh3 value as `favicon-hash`, and expose it in the script output.
* Ensure consistent Nuclei script results by reusing the formatter across active data merging and DB storage paths.

**Final goal**
Mimic Shodan’s `http.favicon.hash` filter.

**Next step**
Add favicon-related filtering keyword support (CLI + Web).

**Sample:**
image

### DIFF
--- a/ivre/db/__init__.py
+++ b/ivre/db/__init__.py
@@ -56,6 +56,7 @@ from ivre import config, flow, nmapout, passive, utils, xmlnmap
 from ivre.active.cpe import add_cpe_values
 from ivre.active.data import (
     add_hostname,
+    format_nuclei_output,
     handle_http_content,
     handle_http_headers,
     handle_tlsx_result,
@@ -3250,6 +3251,7 @@ class DBNmap(DBActive):
                     name += " (%s)" % rec["matcher_name"]
                     nuclei_data["matcher-name"] = rec["matcher_name"]
                 nuclei_data["name"] = name
+                extracted_results = None
                 for key in [
                     "curl-command",
                     "extracted-results",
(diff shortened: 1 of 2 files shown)


ivre/db/__init__.py
```diff
@@ -3260,8 +3262,12 @@ def store_scan_json_nuclei(
                 ]:
                     if key in rec:
                         nuclei_data[key] = rec[key]
+                        if key == "extracted-results":
+                            extracted_results = rec[key]
```
p-l-:
```suggestion
```

ivre/db/__init__.py
```diff
@@ -3260,8 +3262,12 @@ def store_scan_json_nuclei(
...
                     elif (alt_key := key.replace("-", "_")) in rec:
                         nuclei_data[key] = rec[alt_key]
+                        if key == "extracted-results":
+                            extracted_results = rec[alt_key]
```
p-l-:
```suggestion
```

ivre/db/__init__.py
```diff
@@ -3275,10 +3281,17 @@ def store_scan_json_nuclei(
                 ]:
                     if key in rec:
                         nuclei_data[key] = rec[key]
+                if rec.get("template") == "favicon-detect" and extracted_results:
+                    if isinstance(extracted_results, (list, tuple)):
+                        favhash = extracted_results[0] if extracted_results else None
+                    else:
+                        favhash = extracted_results
+                    if favhash:
+                        nuclei_data["favicon-hash"] = favhash
```
p-l-:
```suggestion
```

p-l- (commented):
Hi! Thanks again for your contributions!
I like the idea of having a specific output, but I'm more reluctant to accept a specific format here, since we could end up with one key / value per template of interest.
I like that Nuclei, unlike Nmap, has a generic format to store the data it collects, regardless of the template, and would prefer to keep that format.
However, depending on why you would need that formatted output, you can use the template name and the existing extracted_results list.
For example, if you wanted to implement a `.topvalues()` pseudo-field, it should be easy (I can help you if needed) to implement.
To implement a filter, you can already use `.searchscript()`:
```python
db.view.searchscript(
    "http-nuclei",
    values={
        "template": "favicon-detection",
        "extracted_results": mmh3_value,
    },
)`
```
Let me know what you think.

PR #1896, opened 2026-06-08
feat(web-ui): add pagination controls to /ui/view

Ch4mpa9ne:
## Problem

The legacy Angular WebUI supports result navigation through `skip` / `limit`.

The new React `/ui/view` reads `limit`, but always queries the first result window, so users can only browse the first page of View results.

## Solution

This PR adds pagination controls to `/ui/view`.

The route keeps `skip` / `limit` in the URL for shareable navigation, then translates them into IVRE's existing query syntax (`limit:N skip:N`) when fetching View hosts. The count query continues to use the unpaginated filter so the UI can render correct page bounds.

The change is intentionally scoped to the View section for now. The helper logic is small enough to reuse later, but Active and other sections keep their existing behavior.

Covered behavior includes:

- reading `skip` from URL search params for `/ui/view`
- preserving existing filter and `limit` state
- rendering first / previous / next / last pagination controls
- resetting `skip` when filters change
- keeping `/ui/active` unaffected

## Validation

- [x] pnpm typecheck
- [x] pnpm lint
- [x] pnpm test
- [x] pnpm build
- [x] ./pkg/buildwebui
- [x] manually tested against a live IVRE backend with more than one page of results

I kept the change focused on `/ui/view`; happy to adjust the pagination UI or behavior if you prefer a different approach.


p-l-:
Thanks @Ch4mpa9ne! I have merged your commit together with fixes & more tests in #1898.

PR #1912, opened 2026-06-28
fix: preserve backslash characters in regex filter input

### LINKED ISSUE #1486: Escape character not working while using regular expressions in web interface
### IVRE Version

Copyright 2011 - 2022 Pierre LALET <pierre@droids-corp.org>
Version 0.9.18.dev124-dockerhub

Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110]

Linux 62dce771524b 5.4.0-136-generic #153-Ubuntu SMP Thu Nov 24 15:56:58 UTC 2022 x86_64

Dependencies:
    Python module pymongo: 3.11.0
    Python module sqlalchemy: missing
    Python module psycopg2: missing
    Python module krbV: missing
    Python module pycurl: missing
    Python module PIL: missing
    Python module MySQLdb: missing
    Python module dbus: missing
    Python module matplotlib: missing
    Python module bottle: 0.12.19
    Python module OpenSSL: 20.0.1
    Python module tinydb: missing
### Summary

When using regular expressions ( in search, or cert.subject ) in the web GUI special characters cannot be escaped. E.g: searching for wildcard characters in SSL certificate subjects using '\' as an escaping character ( in previous example something like: cert.subject:/\.\*DOMAIN.NET/ ) will result in removing the backslash. Results will be as if the backslash never existed and thus a different result will be shown than intended

### Expected behavior

Escaping special characters (.*) with an escape character (backslash) while using regular expressions will lead to passing these characters to the engine and a search will be performed on the search string including these characters.

### Actual behavior

Regular Expression escape character (backslash) is removed and therefore not able to escape characters that need to be part of the search string

### How to reproduce

Use a regular expression in a filter / search using a backslash to escape characters like . or *

aishsidya0402-netizen:
## What does this PR do?
Fixes backslash escape characters being silently stripped from regex 
filter input in the web interface. Users can now search using patterns 
like `hostname:/\.google\./` and the backslash is correctly passed to 
the backend.

## Why was this PR needed?
Issue #1486 reported that backslash characters typed into the web UI 
filter bar are removed before the query reaches the search engine. 
Investigation of `web/static/ivre/params.js` revealed that in the 
`parse_params` state machine, the backslash character was being added 
to `curtokenprotected` (display only) but not to `curtoken` (the 
value sent to the API) in states 0, 1, and 2.

## What are the relevant issue numbers?
Closes #1486

## Does this PR meet the acceptance criteria?
- [x] Tests added for new/changed behavior
- [x] All tests passing
- [ ] Follows project style guide
- [x] No breaking changes introduced

### DIFF
--- a/tests/test_params_regex.js
+++ b/tests/test_params_regex.js
@@ -1,0 +1,68 @@
+// Test: backslash characters are preserved in regex filter input
+// Tests the parse_params fix in web/static/ivre/params.js
+
+function parse_params_curtoken(query) {
+    var state = 0;
+    var curtoken = "";
+    var curchar;
+    for (var i = 0; i < query.length; i++) {
+        curchar = query[i];
+        switch (state) {
+            case 0:
+                switch (curchar) {
+                    case " ": break;
+                    case "\\":
+                        state = 3;
+                        curtoken += curchar;
+                        break;
+                    default:
+                        curtoken = curchar;
+                        state = 1;
+                }
+                break;
+            case 1:
+                switch (curchar) {
+                    case " ": state = 0; break;
+                    case "\\":
+                        state = 3;
+                        curtoken += curchar;
+                        break;
+                    default:
+                        curtoken += curchar;
+                }
+                break;
+            case 3:
+                curtoken += curchar;
+                state = 1;
+                break;
+        }
+    }
+    return curtoken;
+}
+
+var passed = 0;
+var failed = 0;
+
+function assert(description, actual, expected) {
+    if (actual === expected) {
+        console.log("PASS: " + description);
+        passed++;
+    } else {
+        console.log("FAIL: " + description);
+        console.log("  Expected: " + expected);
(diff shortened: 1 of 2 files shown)


aishsidya0402-netizen:
Hi @p-l- — this is my first contribution to ivre. I've fixed issue #1486 by preserving backslash characters in the `parse_params` state machine in `web/static/ivre/params.js`. Would appreciate a review when you have time!

p-l-:
Hi @aishsidya0402-netizen!
As you can see in #1913, the bug was a bit more complex to fix and required changes in the backend. Moreover, IVRE now has a new / future Web UI that also had the same bug that needed to be fixed.
Thanks a lot for your contribution!

aishsidya0402-netizen:
Thank you so much! I learned a lot from this. I really appreciate you taking the time to look into it and implement a more complete fix!

